PR #27, opened 2019-08-09
Added authentication parameters, port

katiecai:
Added user name and password for case where user doesn't want to use localhost, and optional port

PR #28, opened 2019-08-12
Connection string logic and authentication

katiecai:
Adding logic for configuring output/trigger with a connection string instead of individual host/user/pw/port parameters. Needs some cleanup

For the output implementation, I decided to include all params PLUS the connection string as optional attributes and then let the RabbitMQ API handle any missing fields. I think this makes the most sense as I would have to do a lot of unnecessary error checking otherwise.

How to design the trigger is a little trickier. The RabbitMQ API has defaults for any missing fields so the user should essentially be able to pass in whatever parameters they wish to, but this makes the trigger code a lot more complicated if we want to account for all of these combinations. Because I have to create trigger objects, I thought about writing a bunch of overloaded constructors to handle different combinations of filled in/missing fields. I've written a few overloaded constructors in this PR but I wanted some feedback on whether or not this is the best way to handle this issue because then I'd have to create like 10+ constructors. 

Also: closing https://github.com/Azure/azure-functions-rabbitmq-extension/pull/27 to combine authentication and connection string logic into one PR

Edit: Adding requeuing logic here as well since it's not a lot of code. Will need to verify whether this is exactly what we want to do but according to [documentation](https://www.rabbitmq.com/confirms.html#consumer-nacks-requeue) this seems to be the solution we want?

Edit 2: The NACK logic seems to requeue over and over even when something hasn't gone wrong. Some say the default behavior is that the message won't leave the queue if there's an error. I need to confirm this but if it's true we don't need to add any other logic for resending

### DIFF
--- a/src/Config/RabbitMQExtensionConfigProvider.cs
+++ b/src/Config/RabbitMQExtensionConfigProvider.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
         public void ValidateBinding(RabbitMQAttribute attribute, Type type)
         {
             string connectionString = Utility.FirstOrDefault(attribute.ConnectionStringSetting, _options.Value.ConnectionString);
-            string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName) ?? Constants.Localhost;
+            string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName) ?? Constants.LocalHost;
             _logger.LogInformation("Setting hostName to localhost since it was not specified");
             string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName) ?? throw new InvalidOperationException("RabbitMQ queue name is missing");
 
--- a/src/Config/RabbitMQExtensionConfigProvider.cs
+++ b/src/Config/RabbitMQExtensionConfigProvider.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
             string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName);
             string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName) ?? throw new InvalidOperationException("RabbitMQ queue name is missing");
 
-            if (string.IsNullOrEmpty(connectionString) && !hostName.Equals("localhost", StringComparison.InvariantCultureIgnoreCase))
+            if (!string.IsNullOrEmpty(hostName) && string.IsNullOrEmpty(connectionString) && !hostName.Equals("localhost", StringComparison.InvariantCultureIgnoreCase))
             {
                 string userName = Utility.FirstOrDefault(attribute.UserName, _options.Value.UserName) ?? throw new InvalidOperationException("RabbitMQ username is missing");
                 string password = Utility.FirstOrDefault(attribute.Password, _options.Value.Password) ?? throw new InvalidOperationException("RabbitMQ password is missing");
--- a/src/Config/RabbitMQExtensionConfigProvider.cs
+++ b/src/Config/RabbitMQExtensionConfigProvider.cs
@@ -63,6 +63,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
         {
             string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName);
             string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName);
+            string userName = Utility.FirstOrDefault(attribute.UserName, _options.Value.UserName);
+            string password = Utility.FirstOrDefault(attribute.Password, _options.Value.Password);
 
             if (string.IsNullOrEmpty(hostName))
             {
@@ -73,24 +75,46 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
             {
                 throw new InvalidOperationException("RabbitMQ queue name is missing");
             }
+
+            if ((string.IsNullOrEmpty(userName) && hostName != "localhost") ||
+                (string.IsNullOrEmpty(password) && hostName != "localhost"))
+            {
+                throw new InvalidOperationException("RabbitMQ username and password required if not connecting to localhost");
+            }
         }
 
         internal RabbitMQContext CreateContext(RabbitMQAttribute attribute)
         {
-            string hostname = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName);
-            string queuename = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName);
+            string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName);
+            string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName);
             string exchange = Utility.FirstOrDefault(attribute.Exchange, _options.Value.Exchange) ?? string.Empty;
+            string userName = Utility.FirstOrDefault(attribute.UserName, _options.Value.UserName) ?? "guest";
+            string password = Utility.FirstOrDefault(attribute.Password, _options.Value.Password) ?? "guest";
+            int port = 0;
+
+            if (attribute.Port != 0)
+            {
+                port = attribute.Port;
+            }
+            else if (_options.Value.Port != 0)
+            {
+                port = _options.Value.Port;
+            }
+
             IBasicProperties properties = attribute.Properties;
 
             var resolvedAttribute = new RabbitMQAttribute
             {
-                HostName = hostname,
-                QueueName = queuename,
+                HostName = hostName,
+                QueueName = queueName,
                 Exchange = exchange,
+                UserName = userName,
+                Password = password,
+                Port = port,
                 Properties = properties,
             };
 
-            IRabbitMQService service = GetService(hostname, queuename);
+            IRabbitMQService service = GetService(hostName, queueName, userName, password, port);
 
             return new RabbitMQContext
             {
@@ -99,9 +123,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
             };
         }
 
-        internal IRabbitMQService GetService(string hostname, string queuename)
+        internal IRabbitMQService GetService(string hostName, string queueName, string userName, string password, int port)
         {
-            return _rabbitMQServiceFactory.CreateService(hostname, queuename);
+            return _rabbitMQServiceFactory.CreateService(hostName, queueName, userName, password, port);
         }
     }
 }
--- a/src/Config/RabbitMQExtensionConfigProvider.cs
+++ b/src/Config/RabbitMQExtensionConfigProvider.cs
@@ -61,35 +61,25 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 
         public void ValidateBinding(RabbitMQAttribute attribute, Type type)
         {
+            string connectionString = Utility.FirstOrDefault(attribute.ConnectionStringSetting, _options.Value.ConnectionString);
             string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName);
-            string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName);
-            string userName = Utility.FirstOrDefault(attribute.UserName, _options.Value.UserName);
-            string password = Utility.FirstOrDefault(attribute.Password, _options.Value.Password);
+            string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName) ?? throw new InvalidOperationException("RabbitMQ queue name is missing");
 
-            if (string.IsNullOrEmpty(hostName))
+            if (string.IsNullOrEmpty(connectionString) && !hostName.Equals("localhost", StringComparison.InvariantCultureIgnoreCase))
             {
-                throw new InvalidOperationException("RabbitMQ host name is missing");
-            }
-
-            if (string.IsNullOrEmpty(queueName))
-            {
-                throw new InvalidOperationException("RabbitMQ queue name is missing");
-            }
-
-            if ((string.IsNullOrEmpty(userName) && hostName != "localhost") ||
-                (string.IsNullOrEmpty(password) && hostName != "localhost"))
-            {
-                throw new InvalidOperationException("RabbitMQ username and password required if not connecting to localhost");
+                string userName = Utility.FirstOrDefault(attribute.UserName, _options.Value.UserName) ?? throw new InvalidOperationException("RabbitMQ username is missing");
+                string password = Utility.FirstOrDefault(attribute.Password, _options.Value.Password) ?? throw new InvalidOperationException("RabbitMQ password is missing");
             }
         }
 
         internal RabbitMQContext CreateContext(RabbitMQAttribute attribute)
         {
+            string connectionString = Utility.FirstOrDefault(attribute.ConnectionStringSetting, _options.Value.ConnectionString);
             string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName);
             string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName);
             string exchange = Utility.FirstOrDefault(attribute.Exchange, _options.Value.Exchange) ?? string.Empty;
-            string userName = Utility.FirstOrDefault(attribute.UserName, _options.Value.UserName) ?? "guest";
-            string password = Utility.FirstOrDefault(attribute.Password, _options.Value.Password) ?? "guest";
+            string userName = Utility.FirstOrDefault(attribute.UserName, _options.Value.UserName);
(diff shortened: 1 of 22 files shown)


src/RabbitMQAttribute.cs
```diff
@@ -29,6 +29,17 @@ public sealed class RabbitMQAttribute : Attribute
...
+        [AutoResolve]
+        public string Password { get; set; }
+
+        // Optional
+        public int Port { get; set; }
+
+        public string ConnectionStringSetting { get; set; }
+
         [AutoResolve]
         public string Exchange { get; set; }
```
pragnagopa: Is exchange optional? If yes, add documentation in summary tags with info on the default value

src/Trigger/RabbitMQListener.cs
```diff
@@ -59,6 +60,7 @@ public Task StartAsync(CancellationToken cancellationToken)
 
             _consumer.Received += (model, ea) =>
             {
+                _channel.BasicAck(ea.DeliveryTag, true);
```
pragnagopa: is this an intentional change for this PR? Will you be addressing Acks on a separate PR?

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,24 +9,39 @@ namespace Microsoft.Azure.WebJobs
...
-        public RabbitMQTriggerAttribute(string hostName, string queueName, ushort batchNumber)
+        public RabbitMQTriggerAttribute(string hostName, string userName, string password, int port, string queueName)
         {
             this.HostName = hostName;
+            this.UserName = userName;
+            this.Password = password;
+            this.Port = port;
             this.QueueName = queueName;
-            this.BatchNumber = batchNumber;
+            this.BatchNumber = 1;
```
pragnagopa: Can BatchNumber be set by users?

src/Trigger/RabbitMQTriggerAttributeBindingProvider.cs
```diff
@@ -39,13 +39,33 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
+            string userName = Resolve(attribute.UserName);
+
+            string password = Resolve(attribute.Password);
+
+            int port = attribute.Port;
+
             ushort batchNumber = attribute.BatchNumber;
 
-            IRabbitMQService service = _provider.GetService(hostName, queueName);
+            if (string.IsNullOrEmpty(connectionString) && !string.IsNullOrEmpty(hostName) && ((string.IsNullOrEmpty(userName) && hostName != "localhost") ||
```
pragnagopa: defaulting to localhost if hostName is not specified will make this check simpler. Also, move this the condition check to Utility and add Unit tests

test/WebJobs.Extensions.RabbitMQ.Samples/RabbitMQSamples.cs
```diff
@@ -16,7 +16,7 @@ public static class RabbitMQSamples
         public static void TimerTrigger_StringOutput(
             [TimerTrigger("00:01")] TimerInfo timer,
             [RabbitMQ(
-                HostName = "localhost",
+                ConnectionStringSetting = "amqp://guest:guest@localhost:5672/",
```
pragnagopa: do not give connection string here. this needs to be an environment variable

test/WebJobs.Extensions.RabbitMQ.Samples/RabbitMQSamples.cs
```diff
@@ -76,7 +70,7 @@ public static class RabbitMQSamples
 
         // Trigger samples
         public static void RabbitMQTrigger_String(
-             [RabbitMQTrigger("localhost", "queue")] string message,
+             [RabbitMQTrigger("amqp://guest:guest@localhost:5672/", "queue")] string message,
```
pragnagopa: Add test that uses username and password. Note: UserName and password should be setting names as well similar to Twilio SID and password

test/WebJobs.Extensions.RabbitMQ.Samples/RabbitMQSamples.cs
```diff
@@ -85,21 +79,21 @@ public static class RabbitMQSamples
...
-            [RabbitMQTrigger("localhost", "queue")] TestClass pocObj,
+            [RabbitMQTrigger("queue")] TestClass pocObj,
             ILogger logger)
         {
             logger.LogInformation($"RabbitMQ queue trigger function processed message: {pocObj}");
         }
 
         public static void RabbitMQTrigger_RabbitMQOutput(
-            [RabbitMQTrigger("localhost", "queue")] string inputMessage,
+            [RabbitMQTrigger("queue")] string inputMessage,
```
pragnagopa: Add comments to indicate defaults to localhost if HostName is not specified.

src/RabbitMQAttribute.cs
```diff
@@ -29,6 +29,17 @@ public sealed class RabbitMQAttribute : Attribute
...
+        [AutoResolve]
+        public string UserName { get; set; }
+
+        [AutoResolve]
+        public string Password { get; set; }
+
+        // Optional
+        public int Port { get; set; }
+
+        public string ConnectionStringSetting { get; set; }
```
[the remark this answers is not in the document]
pragnagopa: Also, Missing attribute AppSetting on Password and Username

src/RabbitMQAttribute.cs
```diff
@@ -29,6 +29,17 @@ public sealed class RabbitMQAttribute : Attribute
         [AutoResolve]
         public string QueueName { get; set; }
```
pragnagopa: Nit: remove this. references across all the files

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,24 +9,39 @@ namespace Microsoft.Azure.WebJobs
     [Binding]
     public sealed class RabbitMQTriggerAttribute : Attribute
     {
-        public RabbitMQTriggerAttribute(string hostName, string queueName)
+        public RabbitMQTriggerAttribute(string connectionString, string queueName)
+        {
+            this.ConnectionStringSetting = connectionString;
```
pragnagopa: remove this.

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,24 +9,39 @@ namespace Microsoft.Azure.WebJobs
...
             this.HostName = hostName;
+            this.UserName = userName;
+            this.Password = password;
+            this.Port = port;
             this.QueueName = queueName;
-            this.BatchNumber = batchNumber;
+            this.BatchNumber = 1;
         }
 
+        public string ConnectionStringSetting { get;  }
```
pragnagopa: Missing ConnectionString attribute

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,24 +9,39 @@ namespace Microsoft.Azure.WebJobs
...
+
         public string HostName { get; }
 
         public string QueueName { get; }
 
+        public string UserName { get; }
```
pragnagopa: Missing AppSetting attribute on username and password

src/Config/RabbitMQExtensionConfigProvider.cs
```diff
@@ -61,36 +61,54 @@ public void Initialize(ExtensionConfigContext context)
 
         public void ValidateBinding(RabbitMQAttribute attribute, Type type)
         {
+            string connectionString = Utility.FirstOrDefault(attribute.ConnectionStringSetting, _options.Value.ConnectionString);
             string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName);
```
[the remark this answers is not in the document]
pragnagopa: Add new file constants.cs and move localhost to a constant

src/Config/RabbitMQExtensionConfigProvider.cs
```diff
@@ -61,36 +61,54 @@ public void Initialize(ExtensionConfigContext context)
...
-            string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName);
-
-            if (string.IsNullOrEmpty(hostName))
-            {
-                throw new InvalidOperationException("RabbitMQ host name is missing");
-            }
+            string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName) ?? throw new InvalidOperationException("RabbitMQ queue name is missing");
 
-            if (string.IsNullOrEmpty(queueName))
+            if (!string.IsNullOrEmpty(hostName) && string.IsNullOrEmpty(connectionString) && !hostName.Equals("localhost", StringComparison.InvariantCultureIgnoreCase))
```
pragnagopa: Same here: Move the condition to check for localhost to utility and you can reuse that function in TriggerBindingProvider as well

src/Services/RabbitMQService.cs
```diff
@@ -10,15 +10,50 @@ internal sealed class RabbitMQService : IRabbitMQService
...
         {
-            _hostName = hostName ?? throw new ArgumentNullException(nameof(hostName));
+            _connectionString = connectionString;
+            _hostName = hostName;
             _queueName = queueName ?? throw new ArgumentNullException(nameof(queueName));
+            _userName = userName;
+            _password = password;
+            _port = port;
+
+            ConnectionFactory connectionFactory = new ConnectionFactory();
```
pragnagopa: Move logic that builds connectionFactory to a separate method. Add unit tests that verify various if conditions

src/Services/RabbitMQService.cs
```diff
@@ -10,15 +10,50 @@ internal sealed class RabbitMQService : IRabbitMQService
...
+                connectionFactory.Password = _password;
+            }
+
+            if (_port != 0)
+            {
+                connectionFactory.Port = _port;
+            }
 
-            ConnectionFactory connectionFactory = new ConnectionFactory() { HostName = _hostName };
             _channel = connectionFactory.CreateConnection().CreateModel();
```
pragnagopa: Nit: Instead of exposing GetBatch, GetChannel methods, expose getters
```
 public IBasicPublishBatch Batch => _batch;

public IModel Channel => _channel;
```
Change IRabbitMQService to
```
   IModel Channel { get; }

   IBasicPublishBatch Batch { get; }
```

pragnagopa: Also, rename channel to  Model as the description here: https://www.rabbitmq.com/releases/rabbitmq-dotnet-client/v3.4.2/rabbitmq-dotnet-client-3.4.2-client-htmldoc/html/type-RabbitMQ.Client.IModel.html states IModel is interface for AMQP Model

pragnagopa: Rename Batch to BasicPublishBatch. Although BasicPublishBatch is not descriptive, name Batch is misleading

src/Trigger/RabbitMQListener.cs
katiecai: I just included it as a part of this PR because the samples don't work as expected otherwise. But more in depth changes would be made in a separate PR

src/Trigger/RabbitMQTriggerAttribute.cs
katiecai: I used to have an overloaded constructor that takes in batch number but removed it since we don't know how we want to handle batching yet. I can remove anything related to batching from this PR if that makes more sense, since right now we default to one message per batch no matter what

src/RabbitMQAttribute.cs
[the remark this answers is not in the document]
katiecai: Is there an AutoResolve for type int?

src/RabbitMQAttribute.cs
katiecai: Removing exchange for now and setting default value to the empty string

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,24 +9,39 @@ namespace Microsoft.Azure.WebJobs
     [Binding]
     public sealed class RabbitMQTriggerAttribute : Attribute
     {
-        public RabbitMQTriggerAttribute(string hostName, string queueName)
+        public RabbitMQTriggerAttribute(string connectionString, string queueName)
```
pragnagopa: reaname connectionString to connectionStringSetting

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,24 +9,39 @@ namespace Microsoft.Azure.WebJobs
...
-            this.HostName = hostName;
-            this.QueueName = queueName;
-            this.BatchNumber = batchNumber;
+            QueueName = queueName;
         }
 
+        public RabbitMQTriggerAttribute(string hostName, string userName, string password, int port, string queueName)
+        {
+            HostName = hostName;
+            UserName = userName;
```
pragnagopa: Same for username and password: raname to UserNameSetting, PasswordSetting

src/Utility.cs
```diff
@@ -11,5 +12,28 @@ internal static string FirstOrDefault(params string[] values)
...
+                    return true;
+                }
+
+                return false;
+            });
+        }
+
+        internal static bool ValidateUserNamePassword(string userName, string password, string hostName)
+        {
+            if (!hostName.Equals("localhost", StringComparison.InvariantCultureIgnoreCase) && (string.IsNullOrEmpty(userName) || string.IsNullOrEmpty(password)))
```
pragnagopa: change references for "localhost" to Constants.LocalHost

test/WebJobs.Extensions.RabbitMQ.Tests/UtilityTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Text;
+using Microsoft.Azure.WebJobs.Extensions;
+using Xunit;
+
+namespace WebJobs.Extensions.RabbitMQ.Tests
+{
+    public class UtilityTests
+    {
+        [Fact]
+        public void InvalidCredentials_ReturnsFalse()
```
pragnagopa: Change this to use theory. Pass expected result as a praram

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQServiceTests.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                Assert.Null(factory.Uri);
+                Assert.Equal("localhost", factory.HostName);
+                Assert.Equal("guest", factory.UserName);
+                Assert.Equal("guest", factory.Password);
+                Assert.Equal(5672, factory.Port);
+            }
+            else if (connectionString != null)
+            {
+                Assert.Equal(new Uri("amqp://user:PASSWORD@52.175.195.81:5672"), factory.Uri);
```
pragnagopa: use dummy ip

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQServiceTests.cs
```diff
@@ -0,0 +1,47 @@
...
+                Assert.Equal("52.175.195.81", factory.HostName);
+                Assert.Equal("user", factory.UserName);
+                Assert.Equal("PASSWORD", factory.Password);
```
pragnagopa: rename PASSWORD to testPassoword and rename user to testUserName

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQServiceTests.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace WebJobs.Extensions.RabbitMQ.Tests
+{
+    public class RabbitMQServiceTests
+    {
+        [Theory]
+        [InlineData(null, "localhost", "queue", "guest", "guest", 5672)]
+        [InlineData("amqp://user:PASSWORD@52.175.195.81:5672", null, "queue", null, null, null)]
+        [InlineData(null, "localhost", "queue", null, null, 0)] // Should fill in "guest", "guest", 5672
+        public void Handles_Connection_Attributes_And_Options(string connectionString, string hostName, string queueName, string userName, string password, int port)
```
pragnagopa: pass expected values as part of params. Yes signature will look ugly, but test code will lot simpler as you just need to check for expected values

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQServiceTests.cs
katiecai: If I do this the connection will fail - what would be the best solution for this

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQServiceTests.cs
pragnagopa: Just test CreateConnectionFactory. do not create RabbitMQ Service

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQServiceTests.cs
pragnagopa: Just test CreateConnectionFactory. do not create RabbitMQ Service

katiecai: Do I mock RabbitMQService? How do I test this without actually creating the connection since that's what the method does?

PR #45, opened 2019-08-28
Add dead letter exchange logic

katiecai:
Adds poisoned messages delivered to the triggered queue to a dead letter exchange with the specified name. We make this attribute a requirement for the trigger binding, which means that the developer must configure their queue with the same dead letter exchange. 

As of now, I'm not making the dead letter exchange name required for the output binding, since the trigger in particular is what has to handle any message delivery issues.

Still trying to figure out how to verify that the poisoned messages are actually in the DLX.

### DIFF
--- a/src/Trigger/RabbitMQTriggerAttribute.cs
+++ b/src/Trigger/RabbitMQTriggerAttribute.cs
@@ -9,24 +9,27 @@ namespace Microsoft.Azure.WebJobs
     [Binding]
     public sealed class RabbitMQTriggerAttribute : Attribute
     {
-        public RabbitMQTriggerAttribute(string connectionStringSetting, string queueName)
+        public RabbitMQTriggerAttribute(string connectionStringSetting, string queueName, string deadLetterExchangeName)
         {
             ConnectionStringSetting = connectionStringSetting;
             QueueName = queueName;
+            DeadLetterExchangeName = deadLetterExchangeName;
         }
 
-        public RabbitMQTriggerAttribute(string queueName)
+        public RabbitMQTriggerAttribute(string queueName, string deadLetterExchangeName)
         {
             QueueName = queueName;
+            DeadLetterExchangeName = deadLetterExchangeName;
         }
 
-        public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string passwordSetting, int port, string queueName)
+        public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string passwordSetting, int port, string queueName, string deadLetterExchangeName)
         {
             HostName = hostName;
             UserNameSetting = userNameSetting;
             PasswordSetting = passwordSetting;
             Port = port;
             QueueName = queueName;
+            DeadLetterExchangeName = deadLetterExchangeName;
         }
 
         [ConnectionString]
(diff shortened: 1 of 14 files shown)


pragnagopa:
We’re you able to verify this e2e?

katiecai:
@pragnagopa Like on app service? Or just through the samples?

pragnagopa:
Yes just through samples!

test/WebJobs.Extensions.RabbitMQ.Samples/RabbitMQSamples.cs
```diff
@@ -106,6 +106,14 @@ public static class RabbitMQSamples
             logger.LogInformation($"RabbitMQ queue trigger function processed message: {pocObj}");
         }
 
+        public static void RabbitMQTrigger_JsonToPOCO(
+            [RabbitMQTrigger(connectionStringSetting: "rabbitMQ", "new_test_queue-poison", "dlxName")] string res,
+            ILogger logger)
+        {
+            logger.LogInformation($"RabbitMQ queue trigger function processed message: {res}");
+        }
+
```
katiecai: So this sample actually doesn't work. :( Headers are automatically added to deadlettered messages so configuring "dlxName" as a header won't match the headers of the poisoned message. Let me know what you think! @pragnagopa

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,24 +9,28 @@ namespace Microsoft.Azure.WebJobs
...
-        public RabbitMQTriggerAttribute(string queueName)
+
+        public RabbitMQTriggerAttribute(string queueName, string deadLetterExchangeName = "")
         {
             QueueName = queueName;
+            DeadLetterExchangeName = deadLetterExchangeName;
         }
 
-        public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string passwordSetting, int port, string queueName)
+        public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string passwordSetting, int port, string queueName, string deadLetterExchangeName = "")
```
pragnagopa: use String.empty instead of  ""

src/Trigger/RabbitMQTriggerAttributeBindingProvider.cs
```diff
@@ -59,6 +59,8 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
 
             string password = Resolve(attribute.PasswordSetting);
 
+            string deadLetterExchangeName = Resolve(attribute.DeadLetterExchangeName) ?? throw new InvalidOperationException("Dead letter exchange name missing");
```
pragnagopa: do not throw if null or empty. Make this optional

test/WebJobs.Extensions.RabbitMQ.Samples/RabbitMQSamples.cs
pragnagopa: Add documentation in the Summary that explains trigger is setup on poison queue from function 'RabbitMQTrigger_JsonToPOCO' , also rename  `RabbitMQTrigger_JsonToPOCO_DeadLetter` to RabbitMQTrigger_Process_PoisonQueue

Issue #46, opened 2019-08-29
Figure out scaling metrics


katiecai:
Currently, the only metrics we use for scaling are queue length. It'd be nice to be able to use time as a metric as well, but here are the issues with that:

- RabbitMQ has no implementation for message timestamp. We could use the [timestamp extension](https://github.com/rabbitmq/rabbitmq-message-timestamp) for this, or we could implement this ourselves (add a timestamp in the header of a message or something). Regardless of which method we were to choose, timestamp would need to be configured on the sending side as well. This complicates things more since it would require the user to add timestamp to their application even if they weren't planning to.
- RabbitMQ has no easy way for you to monitor queues. You can check the number of messages, but not the messages themselves - therefore we wouldn't even be able to check timestamp. From what I found we could do a no ACK (so that the messages stay in the queue even after consumption), but I think that would complicate the ACK protocol that we already have in place. There is also a [management plugin](https://www.rabbitmq.com/management.html) that allows you to monitor messages, but this consumes messages and requeues them. This could result in out-of-order messages in the queue, and I'm not sure we would want to requeue every message.

Lastly, RabbitMQ has a consumer count for queues which I'm not sure would be an important metric for scaling.

@pragnagopa Let me know what you think!

katiecai:
scaling PR in progress: https://github.com/Azure/azure-functions-rabbitmq-extension/pulls

JatinSanghvi:
I think just counting the queue lengths covers all situations. If the average time for processing a message becomes more than the average time interval between messages arriving to the queue, the queue size will start increasing and soon, that will trigger scaling out of the Function app. Let me know if I am not getting something and if there is an example scenario that counters my assumption.

JatinSanghvi:
Closing the issue. We can create a new one if we ever find that the queue length exclusively is not the right choice for scaling decision for any situation.

JatinSanghvi:
Afterthought: Message rate based scaling will be helpful if we have to react quickly to burst of incoming messages, rather than waiting for queue length to hit the threshold.

PR #48, opened 2019-09-03
Add initial scaling ops

katiecai:
Adds VNET scaling logic to the RabbitMQ extension, implementing the new interfaces from https://github.com/Azure/azure-webjobs-sdk/commit/fb7558cedabfa90fda26072c03e4eca26283ecae

Right now we only use queue length as a metric for this trigger. See https://github.com/Azure/azure-functions-rabbitmq-extension/issues/46 for more details

### DIFF
--- a/src/Trigger/RabbitMQListener.cs
+++ b/src/Trigger/RabbitMQListener.cs
@@ -3,18 +3,20 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Azure.WebJobs.Host.Scale;
 using Microsoft.Extensions.Logging;
 using RabbitMQ.Client;
 using RabbitMQ.Client.Events;
 
 namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 {
-    internal sealed class RabbitMQListener : IListener
+    internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTriggerMetrics>
     {
         private readonly ITriggeredFunctionExecutor _executor;
         private readonly string _queueName;
@@ -24,6 +26,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 
         private EventingBasicConsumer _consumer;
         private IRabbitMQModel _rabbitMQModel;
+        private QueueDeclareOk _queueInfo;
         private List<BasicDeliverEventArgs> batchedMessages = new List<BasicDeliverEventArgs>();
 
         private string _consumerTag;
@@ -38,6 +41,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
             _batchNumber = batchNumber;
             _logger = logger;
             _rabbitMQModel = _service.Model;
+            _queueInfo = _service.QueueInfo;
         }
 
         public void Cancel()
@@ -109,7 +113,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
         {
             _rabbitMQModel.BasicAck(ea.DeliveryTag, false);
 
-            if (ea.BasicProperties.Headers == null) {
+            if (ea.BasicProperties.Headers == null)
+            {
                 ea.BasicProperties.Headers = new Dictionary<string, object>();
             }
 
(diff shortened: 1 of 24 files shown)


src/Trigger/RabbitMQListener.cs
```diff
@@ -139,12 +144,77 @@ internal void RepublishMessages(BasicDeliverEventArgs ea)
             }
         }
 
+        public string Id
+        {
+            get
+            {
+                return $"RabbitMQTrigger-{_queueName}".ToLower();
```
mathewc: Recommend adding the FunctionId to this as well. Take a look, at the QueueListener example - it uses {_functionId}-QueueTrigger-{_queue.Name}. FuncitonId represents the individual function the listener is for.

src/Trigger/RabbitMQListener.cs
```diff
@@ -139,12 +144,77 @@ internal void RepublishMessages(BasicDeliverEventArgs ea)
...
+        ScaleStatus IScaleMonitor.GetScaleStatus(ScaleStatusContext context)
+        {
+            ScaleStatus status = new ScaleStatus
+            {
+                Vote = ScaleVote.None,
+            };
+
+            const int NumberOfSamplesToConsider = 5;
+
+            List<RabbitMQTriggerMetrics> metrics = (context as ScaleStatusContext<RabbitMQTriggerMetrics>)?.Metrics?.ToList();
```
mathewc: I realized I had a bug in the QueueListener code where you copied this from. See the latest implementation. Issue was these apis handle both generic/non-generic calls, and there was a casting bug making my impl. not work in some cases. Recommend adding tests for both generic and non-generic for these interfaces as I've done (https://github.com/Azure/azure-webjobs-sdk/pull/2315/files#diff-18dc160d04acf1baa68eee0610eb3d74R101)

src/Trigger/RabbitMQListener.cs
```diff
@@ -139,12 +144,77 @@ internal void RepublishMessages(BasicDeliverEventArgs ea)
...
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await GetMetricsAsync();
+        }
+
+        public async Task<RabbitMQTriggerMetrics> GetMetricsAsync()
+        {
+            return new RabbitMQTriggerMetrics
+            {
+                QueueLength = _queueInfo.MessageCount,
```
mathewc: you need to set Timestamp = DateTime.UtcNow here as well

src/Trigger/RabbitMQListener.cs
```diff
@@ -139,12 +144,77 @@ internal void RepublishMessages(BasicDeliverEventArgs ea)
...
+            }
+
+            bool queueIsIdle = metrics.All(p => p.QueueLength == 0);
+
+            if (queueIsIdle)
+            {
+                status.Vote = ScaleVote.ScaleIn;
+                _logger.LogInformation($"Queue '{_queueInfo.QueueName}' is idle");
+                return status;
+            }
```
mathewc, line 229: Recommend adding the "queue length monotonically increasing" check as well, as in QueueListener

mathewc:
You also need to add tests for this code. See the tests I added in my PR

pragnagopa:
@katiecai  - After addressing CR feedback, please open a new PR against a new 'scaling' branch.

katiecai:
> You also need to add tests for this code. See the tests I added in my PR

Done. I didn't add any log assertions in my unit tests as right now I don't have access to the common testing framework. If there's some way I can get around this/another way I could test for logs that'd be great to know, thanks!

mathewc (approved):
These core changes look good to me. BTW the common test infrastructure is packaged as a nuget library that you can install - other extensions pull this in, e.g. see https://github.com/Azure/azure-functions-servicebus-extension/blob/364bffef4c1f7c017f73a60758170ecbf673d22e/test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/WebJobs.Extensions.ServiceBus.Tests.csproj#L29. Might be good to do that so you can employ the same testing methodology for logs.

src/Trigger/RabbitMQListener.cs
```diff
@@ -30,14 +36,22 @@ internal sealed class RabbitMQListener : IListener
         private bool _disposed;
         private bool _started;
 
-        public RabbitMQListener(ITriggeredFunctionExecutor executor, IRabbitMQService service, string queueName, ushort batchNumber, ILogger logger)
+        public RabbitMQListener(
+            ITriggeredFunctionExecutor executor,
+            IRabbitMQService service,
+            string queueName,
+            ushort batchNumber,
```
mathewc: remove batchNumber as well since not used.

Also, in StartAsync, use `_rabbitMQModel.BasicQos(0, 30, false);` with a TODO comment that these prefetch params should be part of RabbitMQOptions (PrefetchSize, PrefetchCount). If you have time you might just make those changes, taking the options as a param to your listener and using the values from there

src/Trigger/RabbitMQListener.cs
```diff
@@ -140,12 +154,124 @@ internal void RepublishMessages(BasicDeliverEventArgs ea)
...
                 throw new ObjectDisposedException(null);
             }
         }
+
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await GetMetricsAsync();
+        }
+
+        public async Task<RabbitMQTriggerMetrics> GetMetricsAsync()
```
mathewc: remove the async since this function isn't async, and return `return Task.FromResult(metrics);`

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,28 +9,31 @@ namespace Microsoft.Azure.WebJobs
     [Binding]
     public sealed class RabbitMQTriggerAttribute : Attribute
     {
-        public RabbitMQTriggerAttribute(string connectionStringSetting, string queueName, string deadLetterExchangeName = "")
+        public RabbitMQTriggerAttribute(string connectionStringSetting, string queueName, string deadLetterExchangeName = "", ushort batchNumber = 1)
         {
             ConnectionStringSetting = connectionStringSetting;
             QueueName = queueName;
             DeadLetterExchangeName = deadLetterExchangeName;
+            BatchNumber = batchNumber;
```
mathewc: I don't understand BatchNumber - should this even be part of the attribute? Given that you were flowing it incorrectly as a prefetch value

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,28 +9,31 @@ namespace Microsoft.Azure.WebJobs
...
-        public RabbitMQTriggerAttribute(string queueName, string deadLetterExchangeName = "")
+        public RabbitMQTriggerAttribute(string queueName, string deadLetterExchangeName = "", ushort batchNumber = 1)
         {
             QueueName = queueName;
             DeadLetterExchangeName = deadLetterExchangeName;
+            BatchNumber = batchNumber;
         }
 
-        public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string passwordSetting, int port, string queueName, string deadLetterExchangeName = "")
+        public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string passwordSetting, int port, string queueName, string deadLetterExchangeName = "", ushort batchNumber = 1)
```
mathewc: Why are user/password/port separate settings, given that they are part of the connection string? Looks like they're not actually used?

src/Trigger/RabbitMQTriggerAttributeBindingProvider.cs
```diff
@@ -63,13 +63,18 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
+            ushort batchNumber = attribute.BatchNumber;
+
             if (string.IsNullOrEmpty(connectionString) && !Utility.ValidateUserNamePassword(userName, password, hostName))
             {
                 throw new InvalidOperationException("RabbitMQ username and password required if not connecting to localhost");
             }
 
             // If there's no specified batch number, default to 1
-            ushort batchNumber = 1;
+            if (batchNumber == 0)
```
mathewc: recommend removing batchNumber throughout

src/Trigger/RabbitMQTriggerAttribute.cs
katiecai: We wanted to give the user the option to pass in connection info either through a connection string or as separate parameters. That's why the constructor is there, but we haven't been using it much in our testing

katiecai:
Closing as I've migrated the changes over to https://github.com/Azure/azure-functions-rabbitmq-extension/pull/59

PR #50, opened 2019-09-05
Added routing key overload on collector

creyke:
Enabling a common requirement of allowing a RabbitMQ message producer the ability to define a routing key per message, rather than using a queue name.


creyke:
Hmm, the basic interface `IAsyncCollector<in T>` won't support this without a change in the root message type. Closed for now.

PR #57, opened 2019-09-06
making updates to readme

jeffhollan:


jeffhollan:
Moved the wiki to README so more discoverable, added some info.  You can clarify if accurate 😄

PR #59, opened 2019-09-10
Scaling logic

katiecai:
https://github.com/Azure/azure-functions-rabbitmq-extension/pull/48

Some needed changes after testing: https://github.com/Azure/azure-functions-rabbitmq-extension/issues/62

### DIFF
--- a/src/Trigger/RabbitMQTriggerAttribute.cs
+++ b/src/Trigger/RabbitMQTriggerAttribute.cs
@@ -9,23 +9,21 @@ namespace Microsoft.Azure.WebJobs
     [Binding]
     public sealed class RabbitMQTriggerAttribute : Attribute
     {
-        public RabbitMQTriggerAttribute(string connectionStringSetting, string queueName, string deadLetterExchangeName = "", ushort batchNumber = 1)
+        public RabbitMQTriggerAttribute(string connectionStringSetting, string queueName, string deadLetterExchangeName = "")
         {
             ConnectionStringSetting = connectionStringSetting;
             QueueName = queueName;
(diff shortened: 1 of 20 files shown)


src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -9,31 +9,28 @@ namespace Microsoft.Azure.WebJobs
     [Binding]
     public sealed class RabbitMQTriggerAttribute : Attribute
     {
-        public RabbitMQTriggerAttribute(string connectionStringSetting, string queueName, string deadLetterExchangeName = "", ushort batchNumber = 1)
+        public RabbitMQTriggerAttribute(string connectionStringSetting, string queueName, string deadLetterExchangeName = "")
         {
             ConnectionStringSetting = connectionStringSetting;
             QueueName = queueName;
             DeadLetterExchangeName = deadLetterExchangeName;
-            BatchNumber = batchNumber;
```
mathewc: What about the other apparently unused attribute properties like UserName/Password/etc.? Doesn't all this info come from the connection string?

mathewc (approved):

pragnagopa:
👍  merging this

pragnagopa (approved):

src/Trigger/RabbitMQTriggerAttribute.cs
katiecai: Sorry, addressed this on the old PR - we wanted a way for users to configure the connection details using separate attributes as well as the connection string. So far we haven't used it too much in our E2E testing because connection strings are just easier, but initially we wanted both options

Issue #54, opened 2019-09-05
Publish preview package nuget


katiecai:

katiecai:
https://github.com/Azure/azure-functions-rabbitmq-extension/releases/tag/v0.2.2001-beta !!

PR #61, opened 2019-09-10
Update package details

katiecai:
addresses https://github.com/Azure/azure-functions-rabbitmq-extension/issues/54


pragnagopa:
@fabiocav  / @paulbatum  - Here are the predefined license expressions https://spdx.org/licenses/
Do you know if we should use 'MIT` or 'MS-PL` ?
Note: Existing extensions use licence URL  which is now deprecated

fabiocav:
It should be MIT. We need to match all the details we have on other packages (i.e. https://github.com/Azure/azure-webjobs-sdk-extensions/blob/6085a909c2aab7d4ad20cfee9a182f2b0aa10cb3/build/common.props#L16-L26)

pragnagopa (approved):

pragnagopa:
merging this in. Please prepare the release!

Issue #63, opened 2019-09-12
Attribute parameters get serialized incorrectly


jeffhollan:
![image](https://user-images.githubusercontent.com/7132626/64748773-26bd5200-d4c8-11e9-9e0d-4f041f37f6ce.png)

katiecai:
closing with merge of https://github.com/Azure/azure-functions-rabbitmq-extension/pull/64

PR #64, opened 2019-09-12
Only make queue and dlx required in constructor

katiecai:
https://github.com/Azure/azure-functions-rabbitmq-extension/issues/63

Looking at the other extensions, we decided to go with only two overloaded constructors (one that takes in the queue name and one that takes in the queue name + dlx name). Other properties, including the connection string, will then be optional and manually set by the user. Example:

```C#
 public static void RabbitMQTrigger_String(
             [RabbitMQTrigger("new_test_queue", ConnectionStringSetting = "rabbitMQ")] string message,
             string consumerTag,
             ILogger logger)
        {
            logger.LogInformation($"RabbitMQ queue trigger function processed message: {message} and consumer tag: {consumerTag}");
        }
```

### DIFF
--- a/src/Trigger/RabbitMQTriggerAttribute.cs
+++ b/src/Trigger/RabbitMQTriggerAttribute.cs
@@ -21,7 +21,16 @@ namespace Microsoft.Azure.WebJobs
             DeadLetterExchangeName = deadLetterExchangeName;
         }
 
-        public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string passwordSetting, int port, string queueName, string deadLetterExchangeName = "")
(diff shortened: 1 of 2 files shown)


jeffhollan (approved):
Not an expert on the code aspect but the idea / implementation looks great! Thanks Katie

pragnagopa (changes requested):
Fix public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string passwordSetting, int port, string queueName, string deadLetterExchangeName = "") as well. Remove deadLetterExchangeName from the parameters

pragnagopa (approved):
👍

Issue #69, opened 2019-09-26
Support for Header Exchange Trigger


ianrathbone:
Hello, thanks for adding official support for RabbitMQ.

Are there any plans to bring in support for Headers Exchange as a trigger instead of using a queue binding?

ianrathbone:
I've now given this a go with the Pull Request: #71

fabiocav:
FYI @jeffhollan

jeffhollan:
Yep this comes up often.  Appears to be a different "flavor" of messages from RabbitMQ.  Would love to see this added as well as the feature for "durable" queues which is another issue

cachai2:
Needs further investigation

ianrathbone:
Hi, sorry I've not revisited this in some time - would it be preferable to close now given the time that has passed?

PR #71, opened 2019-09-27
Support for the Header Exchange trigger binding

ianrathbone:
One of the concepts supported by RabbitMQ is the "Headers Exchange".
You can read more on this here: https://www.rabbitmq.com/tutorials/amqp-concepts.html

This pull request modifies the existing binding to add support for specifying an exchange instead of a queue. In the RabbitMQ Service this will create a queue based on the exchange name and bind based on specified x-match headers with supplied arguments.

I'm looking to use this in an upcoming project that relies heavily on this methodology, by spinning up several functions that will listen to their own specific headers independently of each other.

I appreciate the work you have done so far and look forward to seeing how I can help extend the functionality of this extension!

### DIFF
--- a/src/Services/RabbitMQService.cs
+++ b/src/Services/RabbitMQService.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Newtonsoft.Json.Linq;
 using RabbitMQ.Client;
 
 namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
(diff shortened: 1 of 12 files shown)


src/Services/RabbitMQService.cs
```diff
@@ -38,6 +39,32 @@ public RabbitMQService(string connectionString, string hostName, string userName
...
+
+            var compiledArguments = new Dictionary<string, object>();
+            if (!string.IsNullOrWhiteSpace(xMatch))
+            {
+                compiledArguments.Add("x-match", xMatch);
+            }
+
+            if (!string.IsNullOrWhiteSpace(arguments))
+            {
+                var argumentObject = JObject.Parse(arguments);
```
ianrathbone, line 56: I shall hold my hands up and confess here that I wasn't sure on the best way to pass in the header arguments in order to pass them to QueueBind as a Dictionary. Any advice here would be appreciated!

fabiocav:
Thank you for sending this! We'll provide some feedback on the PR as soon as possible

@jeffhollan can you please comment on the scenario? The issue is https://github.com/Azure/azure-functions-rabbitmq-extension/issues/69 if you want to move the discussion there.

fabiocav:
Apologies about the lack of response here. 

@jeffhollan / @dariagrigoriu can you please comment on the functionality?

yojagad:
Pinging here as we have been resuming work on this extension. We don't create non-existent queues anymore like we used to, could you rebase and comment on how this functionality might fit in with the existing model ?

Issue #51, opened 2019-09-05
Support custom routing keys per message


creyke:
Please can you make it possible to change the routing key per message. This is a common requirement of RabbitMQ message producers.

Beungoud:
Hello,
To go further on this idea, we also need a way to push messages not only to "queues" but also to rabbitMQ "exchange".
Pushing only to queues seems a limitation on the usage of routing keys.

creyke:
I'm very willing to implement this and open a PR - I just need guidance from the team as to how the meta data of the message should be provided (currently there's just a `byte[]`, so assume there would need to be a parent object of that).

yojagad:
I think the way to go about about this would be, instead of introducing a custom wrapper like in the PR here https://github.com/Azure/azure-functions-rabbitmq-extension/pull/77, actually add support for `IAsyncCollector<BasicDeliverEventArgs>` along with introducing the respective converters for

- Converting from string to BasicDeliverEventArgs (replacing the existing converter from string to byte[])
- Converting from Poco to BasicDeliverEventArgs (replacing the existing converter from Poco to byte[])
- Converting from byte[] to BasicDeliverEventArgs (this would be a new converter to prevent breaking changes)

yojagad:
Other points from further discussion

- As publish method does not take `BasicDeliverEventArgs` directly, we therefore do not have a mechanism to expose this as part of output binding. Reference line https://github.com/Azure/azure-functions-rabbitmq-extension/blob/dev/src/Bindings/RabbitMQAsyncCollector.cs#L33

We will revisit this feature ask when the underlying SDK supports publishing to queue with BasicDeliverEventArgs as input.

cachai2:
Blocked. There is no current API to leverage

PR #77, opened 2019-11-12
Added ability to set a custom routing key per message

creyke:
As discussed in #51, it is an extremely common requirement to set a routing key of a message based on, for example it's content. The current implementation sets the routing key to the `QueueName`.

This PR proposes an abstraction in the generic of the `IAsyncCollector` implementation which enables common meta data related to the message body to be expressed, in this case the `RoutingKey`.

This is an optional property, and is only used if it is explicitly defined, else normal operation resumes.

### DIFF
--- a/src/RabbitMQMessage.cs
+++ b/src/RabbitMQMessage.cs
@@ -1,0 +1,14 @@
+﻿namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
+{
+    public class RabbitMQMessage
+    {
+        public RabbitMQMessage(byte[] body)
+        {
+            Body = body;
+        }
+
+        public byte[] Body { get; set; }
+
+        public string RoutingKey { get; set; }
+    }
+}
--- a/src/Bindings/RabbitMQAsyncCollector.cs
+++ b/src/Bindings/RabbitMQAsyncCollector.cs
@@ -9,7 +9,7 @@ using RabbitMQ.Client;
 
 namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 {
-    internal class RabbitMQAsyncCollector : IAsyncCollector<byte[]>
+    internal class RabbitMQAsyncCollector : IAsyncCollector<RabbitMQMessage>
     {
         private readonly RabbitMQContext _context;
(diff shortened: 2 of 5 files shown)


src/RabbitMQMessage.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
+{
+    public class RabbitMQMessage
+    {
+        public RabbitMQMessage(byte[] body)
```
pragnagopa, line 5: @jeffhollan / @fabiocav - Do you recommend introducing a wrapper type 'RabbitMQMessage'

pragnagopa (commented):
@creyke - Thanks for the PR! Current design adds support to bind to byte[], which makes it easier for existing users to adopt functions. Tagged @jeffhollan and @fabiocav  for further discussion on introducing a wrapper type

src/RabbitMQMessage.cs
creyke: Sorry to chase on this, is it possible to make this adjustment? While I understand this is a API change, I can't imagine many developers could use this library without a meta object, or a way to configure additional details pertaining to and around the message?

creyke:
Sorry to chase on this, is it possible to make this adjustment? While I understand this is a API change, I can't imagine many developers could use this library without a meta object, or a way to configure additional details pertaining to and around the message?

fabiocav:
@creyke apologies for the delayed response. We're reviewing the open PRs that have gone stale (not your fault) in the repo.

Is this still relevant? Do you know if there's a specific reason why the SDK doesn't expose that? (sorry, haven't looked at the details, just assuming it would be needed there as well, unless it's provided as part of a method call).

creyke:
Hi @fabiocav, thanks for picking this up.

I believe this has not been exposed yet due to the most basic of use cases not requiring it.

Anyone using [topic exchanges](https://www.rabbitmq.com/tutorials/amqp-concepts.html#exchange-topic) will require control over the routing key.

If this can come in I'm confident many more customers can move over to this library, and it will not overly complicate the use cases for users which do not need to define it.

PR #84, opened 2020-01-09
Removed the hard coded durability values on QueueDeclare

Adrian10988:
Verified to be working with a local rabbitmq instance and a durable queue. Also validated it blows up if durable = true is specified in the attribute and the queue durability is false

I purposefully didn't support configuration values for this because it would not be known which value to use in case config values and attribute values are used at the same time.

### DIFF
--- a/src/Services/RabbitMQService.cs
+++ b/src/Services/RabbitMQService.cs
@@ -39,7 +39,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
             _model = connectionFactory.CreateConnection().CreateModel();
         }
 
-        public RabbitMQService(string connectionString, string hostName, string queueName, string userName, string password, int port, string deadLetterExchangeName)
+        public RabbitMQService(
+            string connectionString,
+            string hostName,
+            string queueName,
+            string userName,
+            string password,
+            int port,
+            string deadLetterExchangeName,
+            bool isDurable,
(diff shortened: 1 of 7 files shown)


src/Services/RabbitMQService.cs
```diff
@@ -53,15 +62,15 @@ public RabbitMQService(string connectionString, string hostName, string queueNam
             if (!string.IsNullOrEmpty(_deadLetterExchangeName))
             {
                 string deadLetterQueueName = string.Format("{0}-poison", _queueName);
-                _model.QueueDeclare(queue: deadLetterQueueName, durable: false, exclusive: false, autoDelete: false, arguments: null);
+                _model.QueueDeclare(queue: deadLetterQueueName, isDeadLetterExchangeDurable, exclusive: false, autoDelete: false, arguments: null);
```
pragnagopa, line 65: IsDurable?

pragnagopa (commented):
Changes look good. Can you please add a sample here: https://github.com/Azure/azure-functions-rabbitmq-extension/blob/dev/test/WebJobs.Extensions.RabbitMQ.Samples/RabbitMQSamples.cs ?

yojagad:
Closing this as creation of queues within the context of extension isn't supported anymore. Extensions requires a pre-existing and pre-configured queue to work.

PR #86, opened 2020-02-23
Cofiguration extending

paveldayneko:
Que configuration extended. As it would be nice to provide lots of configuration knobs it's expected that it will be done through host.json file. 
In scope of this PR Trigger aka consumer queue configuration added. Model's layer intorduced, it's meant to be closer to service layer and don't mess with Options and Attributes.

From my perspective configuration for trigger and binding should be split as the have only few things in common. Configuration for binding was not implemented, as I don't know what's the maintainer's view on that.

Asyncronous consuming implemented properly using AsyncEventingBasicConsumer.

### DIFF
--- a/src/Bindings/RabbitMQAsyncCollector.cs
+++ b/src/Bindings/RabbitMQAsyncCollector.cs
@@ -30,6 +30,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 
         public Task AddAsync(byte[] message, CancellationToken cancellationToken = default)
         {
+            //TODO: Should we publish into exchange instead?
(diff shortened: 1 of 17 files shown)


paveldayneko:
I would like to contribute to this repo more, and obviously it requires some design decision which I can't make myself, expecting, that they will be accepted

fabiocav:
@paveldayneko apologies for the very delayed response here. Are there specific larger changes you'd like to make? Do you have an issue created for that? We can discuss the approach there.

In the meantime, we can provide guidance on the configuration approach that is consistent with the other extensions, @pragnagopa @mhoeger @ankitkumarr would you be able to help with the review and details on this?

yojagad:
Closing this PR as we don't support queue creation/configuration within the extension. It only takes an existing queue. All configuration/creation need to happen outside the context of the extension.

PR #93, opened 2020-10-15
Fix conflict in queue name in Sample project

Socolin:
There is a conflict in queue names when running all samples at once


yojagad:
Hello, apologies for the delay here. Could you rebase and update this PR, please ?

Socolin:
Since the queue creation / dead letter queue creation have been removed this does not seems to be needed anymore but
the ` RabbitMQTrigger_Process_PoisonQueue`  example is now a duplicate of `RabbitMQTrigger_String` so I pushed a commit to remove this one.

yojagad:
Makes sense. Thanks for the update.

yojagad (approved):

Issue #76, opened 2019-11-01
Make queue declaration optional


asud-solvve:
Current implementation calls QueueDeclare automatically for RabbitMQTriggerAttribute queue.
I want to manage queues manually.
Could you, please, add some option to disable automatic queue declaration?

ThejaChoudary:
@asud-solvve  Aplogies for the late reply.
Is this something your still experiencing with the latest version of Azure.

ThankYou  for the patience

fabiocav:
@jeffhollan FYI for things to track for GA

/cc @pragnagopa

pragnagopa:
@asud-solvve - We do support binding to the rabbitmq client

Here is an example on how to do that:

https://github.com/Azure/azure-functions-rabbitmq-extension/blob/214497c70e28208022c16acf71bf21fb77721da8/test/WebJobs.Extensions.RabbitMQ.Samples/RabbitMQSamples.cs#L79-L87

Is that what you are looking for?

Elthan:
@pragnagopa Is this also possible to do with the RabbitMQTrigger? I want to bind to a queue with the trigger, but I do not have access to create queues, which results in getting rejected.

cachai2:
@cachai2 for tracking

yojagad:
Closing this since we removed the ability to create queue as part of extension that was tracked here #109

PR #94, opened 2020-10-15
Add a way to configure how queues are declared to handle complex scen…

### LINKED ISSUE #68: Error when working with RabbitMQ queue durable
Hi,

I have few queues on RabbitMQ that are set with durable at true. 
With this settings, i get the following error when executing the Azure Function in local:

[9/24/2019 8:35:15 AM] The following 1 functions are in error:
[9/24/2019 8:35:15 AM] TriggerFunction3: Microsoft.Azure.WebJobs.Host: Error indexing method 'TriggerFunction3'. RabbitMQ.Client: The AMQP operation was interrupted: AMQP close-reason, initiated by Peer, code=406, text="PRECONDITION_FAILED - inequivalent arg 'durable' for queue 'test' in vhost '/': received 'false' but current is 'true'", classId=50, methodId=10, cause=.

Please note that if I recreate the same queue without the setting durable. I do not have any problem.
FYI, I am using c# nuget package with version 0.2.2029-beta
Thanks

Socolin:
…ario.

It's now possible to pass an implementation of `IRabbitMQQueueDefinitionFactory`
to a trigger so when the queue is declared, it will use the definition returned
by the methods.

This allow to configure Durable Queue (fixes: #68)
This may help with #76 since it allow to fully configure queue

This allow scenario like in the new Sample: where we may want to have 3 queue
chained together to handle error nicely, when the first queue fail to process
a message, it goes to the second queue, this allow to handle error after all
retry. And if this failed again, it can go to another queue that may be handle
manually and that will contains unexpetected errors.

Queue => Queue-Failed => Queue-Error

Also, I would like to rework `RabbitMQService` to move the logic of how the queue is created somewhere else so I can write unit tests. If you have any indication how this should be done, if you have any preference ?

### DIFF
--- a/src/Config/DefaultRabbitMQServiceFactory.cs
+++ b/src/Config/DefaultRabbitMQServiceFactory.cs
@@ -1,13 +1,24 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.WebJobs.Extensions.RabbitMQ.Trigger;
+
 namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 {
     internal class DefaultRabbitMQServiceFactory : IRabbitMQServiceFactory
     {
-        public IRabbitMQService CreateService(string connectionString, string hostName, string queueName, string userName, string password, int port, string deadLetterExchangeName)
+        public IRabbitMQService CreateService(
+            string connectionString,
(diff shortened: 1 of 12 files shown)


pragnagopa:
@yojagad - is this PR valid as we recently removed support to create non-existing queues?

yojagad:
No, it's not valid anymore. Queue creation and configuration need to happen outside the context of the extension.

PR #102, opened 2020-10-26
[pack] Added build yml and introduced versioning

yojagad:
- Added yml that builds and runs tests. Artifacts are generated only if a PR has [pack] keyword or during CI.
- Artifacts are signed only if it's a release branch.
- Versioning assumes major.minor.patch.build for dev branch and major.minor.patch for release branch(es). Dev branch's version.txt always points to the future version to be merged into release branch and released to myget and/or nuget.


build/version.txt
```diff
@@ -0,0 +1 @@
+1.0.0
```
yojagad: This always points to the next version to be released. After every release, minor version is updated here and merged into dev branch (there will be automation for this).

pragnagopa: Version should be coming from `.csproj` similar to other extensions and webjobs sdk: https://github.com/Azure/azure-webjobs-sdk/pull/2610/files

pragnagopa (commented):
Added a comment

yojagad:
Update where version is got from.

src/WebJobs.Extensions.RabbitMQ.csproj
```diff
@@ -9,7 +9,7 @@
     <SignAssembly>false</SignAssembly>
     <AssemblyOriginatorKeyFile>PublicKey.snk</AssemblyOriginatorKeyFile>
     <DelaySign>true</DelaySign>
-    <Version>0.2.0$(VersionSuffix)-beta</Version>
+    <Version>1.0.0</Version>
```
pragnagopa: This should be `0.2.0$(VersionSuffix)-beta` . `beta` is needed until we GA. `VersionSuffix` is needed to generate packages with build number right?

yojagad: We can probably discuss this more if needed but what I'm thinking is, Dev branch is pointing to a version we are working towards. In this case, when a release is cut off by merging to release branch, 1.0.0 is what goes to nuget. After cutting it off, minor version in Dev branch is updated (1.1.0, 1.2.0, etc) indicating that's the version we are working towards.

All of what I described above would be automated of course.

yojagad: Also, if dev branch needs to generate package, we do it with [pack] and that would generate with build number. Format being major.minor.patch.build (1.0.0.1, 1.0.0.2, etc). Dev packages won't be signed though.

Upgrading any of major, minor or patch version would reset this build number back to 1

pragnagopa: I think we need to consistent across repos with how we version packages. @brettsam might be able to provide better feedback here.

Ideally, using build numbers in the packages directly without us maintaining a counter will help co-relate a package to build number - easy to track which build generated the package.

Regardless - do bring `-beta`  back until we are ready to GA.

fabiocav: We do want to keep the version suffix. Moving to 1.0.0 makes sense, but let's indeed keep the `beta` flag in the meantime.

fabiocav (approved):

yojagad:
I added the -beta prefix. But I'm don't think that adding $(VersionSuffix) is necessary since I already have logic in place for non-release builds to have incremental build numbers here https://github.com/Azure/azure-functions-rabbitmq-extension/blob/3d9ac9ba7adaad4d3c958677b77add4115354db6/build/DetermineVersionAndSigning.ps1 

azure-pipelines.yml in turn uses the versionNumber listed as highlighted here
image

pragnagopa (approved):
Thanks for adding more details. Makes sense. Please make sure artifacts generated are as expected before merging this PR.

PR #107, opened 2020-11-04
Replaced converter with IValueProvider

### LINKED ISSUE #80: Binding to non-json payload does not work with non-dotnet languages
When I try to run the app in Python it throws the following after triggering:

> [11/20/19 7:07:17 PM] Executed 'Functions.RabbitMqTrigger' (Failed, Id=cf468829-43b7-4ec3-9baf-ab44b3e27e17)
[11/20/19 7:07:17 PM] System.Private.CoreLib: Exception while executing function: Functions.RabbitMqTrigger. Microsoft.Azure.WebJobs.Host: Exception binding parameter 'msg'. Newtonsoft.Json: Unexpected character encountered while parsing value: h. Path '', line 0, position 0.

I have a repro of my app here: https://github.com/jeffhollan/repro-python-rabbitmq

Related to #47

yojagad:
Fixes #80 

- Previous implementation worked fine for dotnet but defaulted to JObject for out-of-proc languages
- Updated BindToTrigger but realized that converters were not firing this way. (discussed with @mathewc about this)
- Current workaround removes converters and relies on destinationType to determine whether to convert to POCO or not until we can find a better way.


src/Config/RabbitMQExtensionConfigProvider.cs
```diff
@@ -51,27 +50,23 @@ public void Initialize(ExtensionConfigContext context)
             rule.AddOpenConverter<OpenType.Poco, byte[]>(typeof(PocoToBytesConverter<>));
 
             var triggerRule = context.AddBindingRule<RabbitMQTriggerAttribute>();
-            triggerRule.BindToTrigger<BasicDeliverEventArgs>(new RabbitMQTriggerAttributeBindingProvider(
+
+            // Note that triggerRule.BindToTrigger<BasicDeliverEventArgs> pattern fires up TriggerAdapterBindingProvider which does not respect RequestedType for out-of-proc languages.
```
pragnagopa: nit: Open issue and add link to the issue here

src/Trigger/EventArgsValueProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Newtonsoft.Json;
+using RabbitMQ.Client.Events;
+
+namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
+{
+    public class EventArgsValueProvider : IValueProvider
```
pragnagopa: nit: Rename to `BasicDeliverEventArgsValueProvider`

src/Trigger/EventArgsValueProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                try
+                {
+                    return Task.FromResult(JsonConvert.DeserializeObject(inputValue, Type));
+                }
+                catch (JsonException e)
+                {
+                    // Give useful error if object in queue is not deserialized properly.
+                    string msg = string.Format(@"Binding parameters to complex objects (such as '{0}') uses Json.NET serialization. The JSON parser failed: {1}", Type.Name, e.Message);
+                    throw new InvalidOperationException(msg);
```
pragnagopa: include inner exception?

src/Trigger/RabbitMQTriggerBinding.cs
```diff
@@ -17,16 +18,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
...
-        private readonly ILogger _logger;
 
-        public RabbitMQTriggerBinding(IRabbitMQService service, string hostname, string queueName, ILogger logger)
+        public RabbitMQTriggerBinding(IRabbitMQService service, string hostname, string queueName, ILogger logger, ParameterInfo parameter)
         {
             _service = service;
             _queueName = queueName;
             _hostName = hostname;
             _logger = logger;
+            _parameter = parameter;
```
pragnagopa: instead of passing parameter info,  directly take `Type` ?

pragnagopa (commented):
Added few comments. Please make sure all the samples in `RabbitMQSamples` continue to work on every PR. I do not think they run as part of PR builds

src/Trigger/EventArgsValueProvider.cs
yojagad: Inner exception was blank when I tried it out. Added stack trace instead.

yojagad:
Ran through the samples and they worked alright.

Had to add additional,  special type check for `BasicDeliverEventArgs`.

src/Trigger/EventArgsValueProvider.cs
pragnagopa: ```throw new InvalidOperationException(msg, ex); ```

src/Config/RabbitMQExtensionConfigProvider.cs
```diff
@@ -51,27 +50,22 @@ public void Initialize(ExtensionConfigContext context)
...
-            triggerRule.AddOpenConverter<BasicDeliverEventArgs, OpenType.Poco>(typeof(BasicDeliverEventArgsToPocoConverter<>), _logger);
         }
 
         public void ValidateBinding(RabbitMQAttribute attribute, Type type)
         {
             string connectionString = Utility.FirstOrDefault(attribute.ConnectionStringSetting, _options.Value.ConnectionString);
             string hostName = Utility.FirstOrDefault(attribute.HostName, _options.Value.HostName) ?? Constants.LocalHost;
             _logger.LogInformation("Setting hostName to localhost since it was not specified");
-            string queueName = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName);
+            _ = Utility.FirstOrDefault(attribute.QueueName, _options.Value.QueueName);
```
mathewc: Can this code be removed completely? Looks like the value isn't being used.

src/Config/RabbitMQExtensionConfigProvider.cs
```diff
@@ -51,27 +50,22 @@ public void Initialize(ExtensionConfigContext context)
...
+            // More details about why the BindToTrigger was chosen instead of BindToTrigger<BasicDeliverEventArgs> detailed here https://github.com/Azure/azure-functions-rabbitmq-extension/issues/110
+            triggerRule.BindToTrigger(new RabbitMQTriggerAttributeBindingProvider(
                     _nameResolver,
                     this,
                     _logger,
                     _options,
                     _configuration));
-
-            // Converts BasicDeliverEventArgs to string so user can extract received message.
-            triggerRule.AddConverter<BasicDeliverEventArgs, string>(args => Encoding.UTF8.GetString(args.Body))
```
mathewc, line 62: So you're now implementing your own conversions manually in the value binder, supporting string, BasicDeliverEventArgs, etc. How good is the E2E test coverage for all the various bindings that are supported currently? I'm wondering if these changes mean we're losing some built in conversion handling. E.g. I assume we currently support binding to string, JObject, Pocos, BasicDeliverEventArgs, Stream, TextReader, etc. - basically the built ins defined in ConverterManager https://github.com/Azure/azure-webjobs-sdk/blob/b798412ad74ba97cf2d85487ae8479f277bdd85c/src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs#L33. Need to make sure we're not breaking existing scenarios.

src/Trigger/BasicDeliverEventArgsValueProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                return Task.FromResult<object>(_input);
+            }
+
+            string inputValue = ToInvokeString();
+            if (Type.Equals(typeof(string)))
+            {
+                return Task.FromResult<object>(inputValue);
+            }
+            else
```
mathewc, line 41: This else block is just for Poco handling? How/where is JObject handled?

mathewc:
Note that we've had problems with TriggerAdapterBindingProvider.GetDefaultType in the past. Most recently, @ankitkumarr did some work in https://github.com/Azure/azure-webjobs-sdk/pull/2301. I think the fundamental issue is that we're ignoring requestedType. Note that previously before the changes in the referenced pull request we did use requestedType. Contrast this to [BindToInputBindingProvider](https://github.com/Azure/azure-webjobs-sdk/blob/0dd21b28ae749cc5000339b1d24a4955d39f0395/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToInputBindingProvider.cs#L102) where we do use requestedType. Seems these provider classes are out of sync.  I wonder if in both cases we don't just need to include requestedType in the list of targets we check for conversions for, checking for it first.

src/Config/RabbitMQExtensionConfigProvider.cs
yojagad: I tested this out by binding with the afore mentioned types before and after this change. Binding to types like Stream, TextReader, TextWriter, byte[] did not work in the older version and remains the same with this change as well. Any complex types like that, RabbitMQ expects both the sender and the receiver to know the type-to-byte and byte-to-originalType conversion which is probably where binding to `BasicDeliverEventArgs` comes into picture since it retains the original content as is (I added special check for that as well).

Binding to string, JObject and Poco remains the same before and after this change as well. I verified this. If there's any other scenario I may have missed, please let me know.

src/Trigger/BasicDeliverEventArgsValueProvider.cs
yojagad: Yes. This is handled in dotnet scenario where data is being bound to a Poco like in a below scenario

````
public class TestClass
    {
        public string X { get; set; }
    }

    public static class Function1
    {
        [FunctionName("Function1")]
        public static void Run([RabbitMQTrigger("TestQueue", ConnectionStringSetting = "ConnectionString")] TestClass myQueueItem, ILogger log)
        {
            log.LogInformation($"C# Queue trigger function processed: {myQueueItem}");
        }
    }
````

src/Config/RabbitMQExtensionConfigProvider.cs
yojagad: removed

src/Trigger/EventArgsValueProvider.cs
yojagad: of course x_x

yojagad:
Pinging for reviews please.

pragnagopa (approved):
LGTM. Please wait ACK from @mathewc

ankitkumarr:
@mathewc, yeah I think as you said the `BindToTrigger<T>` is still broken in that it does not honor `requestedType`. I guess the fix in this PR to use the `BindToTrigger` instead of `BindToTrigger<T>` makes sense to unblock RabbitMQ. We should still open a follow-up item in `webjobs-sdk` for the root issue. As any new extension that uses the simplified  `BindToTrigger<T>` would run into the same issue.

yojagad:
Added support for binding to byte[] since this was identified as a gap in previous implementation as well.

yojagad:
Opened an issue here https://github.com/Azure/azure-webjobs-sdk/issues/2636

PR #108, opened 2020-11-05
Don't create non-existent queue and dead letter queue.

### LINKED ISSUE #109: Remove creation of non-existing queue, dead letter queue and exchanges.

yojagad:
Fixes #109

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ The above sample waits on a trigger from the queue named "queue" connected to th
 |Property Name|Description|Example|
 |--|--|--|
 |ConnectionStringSetting|The connection string for the RabbitMQ queue|`amqp://user:password@url:port`|
-|QueueName|The name of the source or destination queue|`myQueue`|
+|QueueName|The name of the source or destination queue. To move failed messages to deadletter queue, please configure queue and exchange(https://www.rabbitmq.com/dlx.html)|`myQueue`|
 |HostName|(optional if using ConnectionStringSetting) Hostname of the queue|`10.26.45.210`|
 |UserName|(optional if using ConnectionStringSetting) User name to access queue|`user`|
 |Password|(optional if using ConnectionStringSetting) Password to access queue|`password1`|
--- a/README.md
+++ b/README.md
@@ -36,7 +36,6 @@ The above sample waits on a trigger from the queue named "queue" connected to th
 |--|--|--|
 |ConnectionStringSetting|The connection string for the RabbitMQ queue|`amqp://user:password@url:port`|
 |QueueName|The name of the source or destination queue|`myQueue`|
-|DeadLetterExchangeName|The name of the deadletter exchange for poison queue patterns|`dtxExchange`|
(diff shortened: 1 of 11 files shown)


pragnagopa (changes requested):
- Please create an issue or link existing issue for this PR
- I believe we should support moving messages to deadletterexchange queue as long as user specifies it. PR should only remove creating queue / dead letter exchange but should allow using existing queues

yojagad:
@pragnagopa Correct me if I'm wrong, but the only action that moves messages to the deadletter exchange and the queue is https://github.com/Azure/azure-functions-rabbitmq-extension/blob/dev/src/Trigger/RabbitMQListener.cs#L178 . As long as the user has them configured correctly and bound the exchange to the queue, that line would either move it to dead letter exchange or they might end up losing the message altogether.

I specifically didn't attempt to do a passive check on whether the exchange or the dead letter queue exists, because there's no reason that the binding itself should fail to come up if either of these aren't configured correctly. Let me know if I'm missing something else though, please.

src/Services/RabbitMQService.cs
```diff
@@ -33,29 +31,13 @@ public RabbitMQService(string connectionString, string hostName, string userName
...
             _rabbitMQModel = new RabbitMQModel(_model);
-
-            _deadLetterExchangeName = deadLetterExchangeName;
             _queueName = queueName ?? throw new ArgumentNullException(nameof(queueName));
 
-            Dictionary<string, object> args = new Dictionary<string, object>();
-
-            // Create dead letter queue
-            if (!string.IsNullOrEmpty(_deadLetterExchangeName))
-            {
```
pragnagopa, line 48: We should allow users to specify a `deadLetterExchangeName` . I think you just have to delete creating the queue here. Also,  log an error if adding message to `deadLetterExchangeName` fails

src/RabbitMQAttribute.cs
```diff
@@ -40,8 +39,5 @@ public sealed class RabbitMQAttribute : Attribute
 
         [ConnectionString]
         public string ConnectionStringSetting { get; set; }
-
-        [AutoResolve]
-        public string DeadLetterExchangeName { get; set; }
```
pragnagopa, line 45: Without this users will not be able to specify an existing queue for `DeadLetterExchangeName` ?

README.md
```diff
@@ -36,7 +36,6 @@ The above sample waits on a trigger from the queue named "queue" connected to th
 |--|--|--|
 |ConnectionStringSetting|The connection string for the RabbitMQ queue|`amqp://user:password@url:port`|
 |QueueName|The name of the source or destination queue|`myQueue`|
```
pragnagopa: nit:  Add `To move failed messages to deadletterqueue, configure queue (add link)`

pragnagopa (approved):
Please run all the samples before merging the PR

yojagad:
Ran through the samples and didn't find issues.

PR #111, opened 2020-11-09
Java support

### LINKED ISSUE #60: Add support for Java

yojagad:
Part of work for #60 

@jeffhollan / @fabiocav  Following the signalr way of adding annotations for an extension (https://github.com/Azure/azure-functions-signalrservice-extension/tree/master/binding-library/java)

cc: @cachai2

Below is an example java function
````
/**
 * Azure Functions with HTTP Trigger.
 */
public class Function {
    @FunctionName("RabbitMQExample")
    public void run(
            @RabbitMQTrigger(
                connectionStringSetting = "ConnectionString",
                queueName = "TestQueue"
            ) String input,
            @RabbitMQOutput(
                connectionStringSetting = "ConnectionString",
                queueName = "hello"
            ) OutputBinding<String> output,
            final ExecutionContext context) {
        context.getLogger().info("Java HTTP trigger processed a request." + input);
        output.setValue(input);
    }
}
````

### CODE REVIEW

**Comment by fabiocav** [comment]
Thanks! Would be good to get @pragnagopa 's feedback here, as I know this differs from the initial proposed implementation

**Comment by pragnagopa** [comment]
@jeffhollan / @fabiocav - I recommend following similar approach as signalr and create Java annotations in this repo.

This allows keeping annotations and c# attributes in sync and can be released independently of azure-function-java-library

File: binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQOutput.java [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+
+import com.microsoft.azure.functions.annotation.CustomBinding;
+
+/**
+ * <p>Place this on a parameter whose value should be set in order for an event to be enqueued to a rabbitMQ queue.
+ * The parameter can be one of the following</p>
+ *
+ * <ul>
+ *     <li>A String java native type</li>
+ *     <li>Any POJO type</li>
```
pragnagopa, line 22: Basic types are supported like any other out-of-proc languange? string, byte[], integer etc?

binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQTrigger.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * <p>Place this on a parameter whose value would come from RabbitMQ and causing the method to run when a new event arrives.
+ * The parameter can be one of the following</p>
+ *
+ * <ul>
+ *     <li>A String java native type</li>
+ *     <li>Any POJO type</li>
+ * </ul>
+ */
```
pragnagopa, line 38: nit: same here. Other basic types are supported as well.

binding-library/java/src/test/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQTriggerTests.java
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
```
pragnagopa: move tests to a folder above? not in the src folder?

pragnagopa (commented):
Added minor comments
- Open issue to add integration test that verifies function.json generated by mvn
- Update PR description with same java function that uses this annotation. Later can be re-used in the documentation

binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQOutput.java
yojagad: I don't think directly binding to byte[], int works for even dotnet even today. 

My understanding was that for in-proc, it would be string, POCO or BasicDeliverEventArgs and for out-of-proc, they would be normal strings or stringified json (except in java, json can be converted to the appropriate POJO as well)

binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQTrigger.java
yojagad: I don't think directly binding to byte[], int works for even for dotnet today for trigger types. 

My understanding was that for in-proc, it would be string, POCO or BasicDeliverEventArgs and for out-of-proc, they would be normal strings or stringified json (except in java, json can be converted to the appropriate POJO as well). Am I missing something else ?

pragnagopa: in dotnet function, user should be able to bind to byte[] as that is the type of the body event is is delivered. If not, please fix this.

For java - as its a typed language - users can request any type - java worker is responsible for converting input data to the type requested in the method signature.

binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQOutput.java
yojagad: I'm only seeing converters for string-to-byte[] and for POCO-to-byte[] for output binding. How could other basic types be supported without the necessary converters ? I guess java worker takes care of it ?

binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQTrigger.java
yojagad: Ah, okay. But java worker doesn't inherently know to convert BasicDeliverEventArgs to primitive types, which means it relies on converters provided by the extension, right ?

The body of event is delivered as byte[] but wrapped as BasicDeliverEventArgs. I was assuming that's why binding to BasicDeliverEventArgs was introduced. I think this probably means a gap that needs to be addressed where we support binding to both byte[] and BasicDeliverEventArgs, right ?

binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQOutput.java
pragnagopa: Here is how it is handled in Java: https://github.com/Azure/azure-functions-java-worker/blob/dev/src/main/java/com/microsoft/azure/functions/worker/binding/RpcIntegerDataSource.java 

For C#, this is the scenario I can think of if user adds a message to rabbitmq converting int to bytes : `byte[] intBytes = BitConverter.GetBytes(intValue);`  ,  trigger binding should support ` [RabbitMQTrigger("test_queue", ConnectionStringSetting = "rabbitMQ")] int res,` ?

yojagad:
@pragnagopa Added issue here https://github.com/Azure/azure-functions-rabbitmq-extension/issues/112

binding-library/java/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+    <version>1.0.0-beta</version>
+    <packaging>jar</packaging>
+    <parent>
+		<groupId>com.microsoft.maven</groupId>
+		<artifactId>java-8-parent</artifactId>
+		<version>8.0.1</version>
+	</parent>
+
+    <name>Microsoft Azure Functions Java RabbitMQ types</name>
+    <description>This package contains all Java interfaces and annotations to interact with Microsoft Azure functions runtime for RabbitMQ Service.</description>
```
pragnagopa: nit: `Microsoft Azure **F**unctions`

binding-library/java/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+    <!--<url> URL to docs to be added</url>-->
```
pragnagopa: nit: delete commented line

binding-library/java/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+    <organization>
```
pragnagopa: delete organization ? I do not see this java library

binding-library/java/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+    </repositories>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.microsoft.azure.functions</groupId>
+            <artifactId>azure-functions-java-library</artifactId>
+            <version>${azure.functions.java.library.version}</version>
+        </dependency>
+        <dependency>
+			    <groupId>junit</groupId>
```
pragnagopa: nit: format

binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQOutput.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.microsoft.azure.functions.annotation.CustomBinding;
+
+/**
+ * <p>Place this on a parameter whose value should be set in order for an event to be enqueued to a rabbitMQ queue.
```
pragnagopa, line 17: Can you please follow format here: https://github.com/Azure/azure-functions-java-library/blob/fc75a74201473af31421420a6193581a6ab22501/src/main/java/com/microsoft/azure/functions/annotation/BlobInput.java#L14-L40 and include example in the comments?

pragnagopa (commented):
Added few more nits . Thanks

binding-library/java/pom.xml
yojagad: I see this in signalr though https://github.com/Azure/azure-functions-signalrservice-extension/blob/master/binding-library/java/pom.xml

pragnagopa: I checked https://github.com/microsoft/azure-maven-plugins/blob/develop/pom.xml as well. lets remove it as it is pointing to a generic Azure URL not specific to this repo

binding-library/java/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+    </repositories>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.microsoft.azure.functions</groupId>
+            <artifactId>azure-functions-java-library</artifactId>
+            <version>${azure.functions.java.library.version}</version>
+        </dependency>
+        <dependency>
+			<groupId>junit</groupId>
```
pragnagopa: fomatting still looks off?

pragnagopa (approved):
Approved. Nit, fix format before merging.

binding-library/java/pom.xml
yojagad: yeah, looks like I was still using tabs in some places resulting in the format

pragnagopa (approved):

Issue #98, opened 2020-10-15
Possible message loss on retries


jeffhollan:
We should move the ACK until after the message is re-published
https://github.com/Azure/azure-functions-rabbitmq-extension/blob/dev/src/Trigger/RabbitMQListener.cs#L15

yojagad:
Fixed by https://github.com/Azure/azure-functions-rabbitmq-extension/pull/114

Issue #66, opened 2019-09-13
Prefetch + Other Options Configurable


mathewc:
Currently we have prefetch count hard coded. We need to explore optimal values for this and come up with a good default, and also make this (and any other important knobs) configurable in host.json.

jeffhollan:
The two properties that are important to surface are number of samples and the target queue length (in addition to what @mathewc called out)

PR #114, opened 2020-11-12
Added configurable options and fixed ack logic

### LINKED ISSUE #99: Allow configurable concurrency similar to Azure Service Bus
https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-output?tabs=csharp#hostjson-settings

Ideally we have a default concurrency of executions happening.  Azure Queues host.json is another example where the underlying provider doesn't have this, but are able to mirror similar functionality

yojagad:
Fixes #99 , #98. 

cc: @cachai2


src/Config/PrefetchOptions.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
+{
+    public class PrefetchOptions
+    {
+        public ushort PrefetchSize { get; set; }
```
pragnagopa: nit: add descriptive comments for these properties.

src/Config/RabbitMQOptions.cs
```diff
@@ -19,5 +16,9 @@ public class RabbitMQOptions
         public string ConnectionString { get; set; }
 
         public int Port { get; set; }
+
+        public PrefetchOptions PrefetchOptions { get; set; }
```
pragnagopa: nit: would be be great if you can fill in comments here too as these are public properties

src/Config/ScaleOptions.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
+{
+    public class ScaleOptions
+    {
+        public int SampleCount { get; set; }
+
```
pragnagopa: nit: same here :) comments

src/Trigger/RabbitMQListener.cs
```diff
@@ -50,6 +55,10 @@ internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTrigge
             _functionDescriptor = functionDescriptor ?? throw new ArgumentNullException(nameof(functionDescriptor));
             _functionId = functionDescriptor.Id;
             _scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{_functionId}-RabbitMQTrigger-{_queueName}".ToLower());
+            _sampleCount = options.Value.ScaleOptions?.SampleCount ?? 5;
+            _targetQueueLength = options.Value.ScaleOptions?.TargetQueueLength ?? 1000;
+            _prefetchCount = options.Value.PrefetchOptions?.PrefetchCount ?? 30;
```
pragnagopa: These defaults need to documented. Please open a doc issue and link here to ensure we do not forget updating docs.

pragnagopa: Also, as @mathewc mentioned in https://github.com/Azure/azure-functions-rabbitmq-extension/issues/66 , it would be good collect data on optimum values that need to be set for prefetch count. Please open a separate issue to run some tests and adjust these values if needed.

src/Trigger/RabbitMQListener.cs
```diff
@@ -221,16 +228,16 @@ private ScaleStatus GetScaleStatusCore(int workerCount, RabbitMQTriggerMetrics[]
                 Vote = ScaleVote.None,
             };
 
-            const int NumberOfSamplesToConsider = 5;
+            int numberOfSamplesToConsider = _sampleCount;
 
-            if (metrics == null || metrics.Length < NumberOfSamplesToConsider)
+            if (metrics == null || metrics.Length < numberOfSamplesToConsider)
```
pragnagopa, line 228: Delete numberOfSamplesToConsider  and use `_sampleCount` directly?

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQListenerTests.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.Azure.WebJobs.Host.Protocols;
 using Microsoft.Azure.WebJobs.Host.Scale;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
```
pragnagopa, line 9: 1. Add unit tests that verifies defaults
2. Add unit test that verifies values overridden
3. Not sure if we have a Samples E2E that show how to configure RabbitMq options - please do add or update sample that configures these values.

pragnagopa (commented):
Added minor comments. Thanks.

src/Config/PrefetchOptions.cs
```diff
@@ -0,0 +1,12 @@
...
+
+        public ushort PrefetchCount { get; set; }
```
mathewc: I think you should make these ints - the rest of our integral config are int

mathewc: Also, are there any limits that need to be applied to these? What are the limits RabbitMQ enforces, and what happens if someone configures over those limits?

yojagad: RabbitMQ client accepts only ushort/uint though..

yojagad: And weirdly I haven't found official docs specifying any limits on them :( 
https://www.cloudamqp.com/blog/2017-12-29-part1-rabbitmq-best-practice.html
https://www.rabbitmq.com/consumer-prefetch.html
I can try to see if I can dig up some more, but I don't think we should enforce something that isn't officially documented.

src/Config/ScaleOptions.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
+{
+    public class ScaleOptions
```
mathewc: This is for runtime driven scale only, right? I think before we do this, we want to have a consistent story for this configuration across all the extensions. I don't believe this is configurable for any of the other extensions currently. Also we'll need to make it clear to users that this is only for RTS and won't affect Consumption scaling, since the ScaleController doesn't currently allow configuration of these. That could be a source of confusion for people. @paulbatum FYI. Let's make any changes to RTS config separate from this PR.

src/Config/PrefetchOptions.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
+{
+    public class PrefetchOptions
```
mathewc: Let's make RabbitMQOptions implement IOptionsFormatter which all our other extension options classes implement, so the configured options are logged. That will help us diagnose issues, particularly now that with your changes the options allow for more interesting behavior changes.

src/Config/ScaleOptions.cs
yojagad: Okay, I'll go ahead and revert this. And I'll open a separate issue for RTS config.

yojagad: Reverted this.

src/Trigger/RabbitMQListener.cs
yojagad: Brought back the variables due to the discussion below with Mathew.

src/Trigger/RabbitMQListener.cs
yojagad: Opened issues https://github.com/Azure/azure-functions-rabbitmq-extension/issues/116 and https://github.com/Azure/azure-functions-rabbitmq-extension/issues/117

src/Config/ScaleOptions.cs
yojagad: Opened issue here https://github.com/Azure/azure-functions-rabbitmq-extension/issues/118

yojagad:
While doing additional testing, I realized that there was no point in implementing configurable prefetch_size since anything other than 0 is not accepted by RabbitMQ and fails with this error https://github.com/streadway/amqp/issues/438 :\ . I'll update this PR accordingly.

yojagad:
Upon further testing, realized that setting prefetchSize to anything other than 0 doesn't seem to work . Encountered the below error :\
image
It could be a bug in the rabbitMQ dotnet client or some other rabbitMQ nuget it's referring as I noticed various other threads complaining about this. Upgraded client to 5.2.0 as well and still encountered the same.

I'll probably remove the option to make prefetchSize configurable.

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQListenerTests.cs
yojagad: Instead of adding it in Samples, I added another test that uses webJobsBuilder with the right PrefetchCount and verifies that the setting returns the same value.

mathewc (approved):

pragnagopa (approved):
💯

freddymk87:
Hello @yojagad  I am using the RabbitMQ binding ,with Azure Functions, c#. 

I want to set the prefetch count to 1000 how can I ? should it be in the Startup function ?

yojagad:
Hello @freddymk87 . The prefetch count is something you should set within host.json. For rabbitMQ with custom prefetch count, the host.json should look like below. 

````
{
    "version": "2.0",
    "extensions": {
        "rabbitMq": {
            "prefetchCount": 1000
         }
    }
}
````

Please ensure you use `1.0.0-beta` version of the extension for the above to work.

freddymk87:
Thank you so much appreciate it

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #115, opened 2020-11-16
[pack] Added build steps for mvn

yojagad:
Building and publishing packages as described in doc. Below is copy paste from one note

Java to maven central (via oss.sonatype.org )
Files: The release job expects the following set of files:

<library>.pom – Renamed version of pom.xml that was used to build the library
<library>.jar - Must always provide for library projects.
<library>-javadoc.jar - Must always provide for library projects.
<library>-sources.jar - Must always provide for library projects.
where <library> is the standard library name and version that maven projects produce for the jar files.

Sample build with both nuget and java : https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=9287&view=artifacts&type=publishedArtifacts

### DIFF
--- a/binding-library/java/pom.xml
+++ b/binding-library/java/pom.xml
@@ -126,6 +126,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>${maven-javadoc.version}</version>
+                <configuration>
+                    <doclint>none</doclint>
+                </configuration>
                 <executions>
                     <execution>
                         <id>attach-javadocs</id>
--- a/binding-library/java/pom.xml
+++ b/binding-library/java/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.microsoft.azure.functions</groupId>
     <artifactId>azure-functions-java-library-rabbitmq</artifactId>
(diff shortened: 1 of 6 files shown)


binding-library/java/pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.microsoft.azure.functions</groupId>
     <artifactId>azure-functions-java-library-rabbitmq</artifactId>
-    <version>1.0.0-beta</version>
+    <version>1.0.0-SNAPSHOT</version>
```
yojagad, line 9: Changed this to snapshot because document was only describing SNAPSHOT and other packages.

binding-library/java/pom.xml
```diff
@@ -126,6 +126,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>${maven-javadoc.version}</version>
+                <configuration>
+                    <doclint>none</doclint>
+                </configuration>
```
yojagad, line 131: javadoc was throwing error otherwise because it was thinking that the `<String>` in `OutputBinding<String>` is an incomplete tag.

pragnagopa (approved):
This is great. Thanks!

PR #147, opened 2020-12-27
Fix System.ObjectDisposedException when calling StopAsync()

Socolin:
Hello

There is a crash when stopping the webjob.

I use this in tests to start/stop some services and when it crash it prevent other services to be disposed / stopped gracefully, and the test process is stuck

```
            var host = builder.Build();
            using (host)
            {
                await host.StartAsync();
                await Task.Delay(2_000);
                await host.StopAsync();
            }
```

### DIFF
--- a/src/Trigger/RabbitMQListener.cs
+++ b/src/Trigger/RabbitMQListener.cs
@@ -81,6 +81,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 
         public void Cancel()
         {
+            if (!_started)
+            {
+                return;
+            }
+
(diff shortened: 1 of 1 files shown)


JatinSanghvi:
Letting the PR go through. A better fix might be possible but I cannot think of it right away. This one does the job, for example, `Cancel` closes the connection if `StopAsync` does not get called before it, and it skips (since there's nothing to be done) if `StopAsync` is already called.

JatinSanghvi (approved):

JatinSanghvi:
Thanks @Socolin 🎉

PR #149, opened 2021-02-13
Caching IModel instead of opening one per invocation

### LINKED ISSUE #148: Many connections opened for IModel
I'm using the code below, based on the examples in the wiki. It is an Azure Function V3 function, using .NET Core 3.1. I'm running this locally.

Whenever I run this, I notice that I have a lot of open connections, up to a point where I'm running out of connections and the application stalls. The RabbitMQ Management Interface shows all these connections as well, and they all disappear when I kill the application. To my understanding, channels are multiplexed on a single connection, so I would assume a single connection, instead of a connection per function invocation. Therefore, I think this is a bug.

As a 'work-around', I now inject a my own `IConnection` in my function class, then use `using (var channel = connection.CreateModel() { .. }` which works as expected. I also notice that it is faster: I was managing 5 msg/sec, now 25 msg/sec.

Code:

```csharp
[FunctionName("FunctionName")]
public async Task RunAsync(
    [RabbitMQTrigger(
        "QueueName",
        ConnectionStringSetting = "ConnectionString")]
    BasicDeliverEventArgs message,
    [RabbitMQ(ConnectionStringSetting = "ConnectionString")]
    IModel channel)
{
    // Some processing of the message here.
    // ...

    // Publish updates to an exchange
    var batch = channel.CreateBasicPublishBatch();

    foreach (var update in updates.Values)
    {
        var (body, properties) = MessagingHelper.ToMessage(update, message);

        batch.Add(
            exchange: "ExchangeName",
            routingKey: "",
            mandatory: false,
            properties: properties,
            body: body);
    }

    batch.Publish();

    // Doesn't help
    // channel.Close()
}
```

yojagad:
Fixes #148

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,3 +1,3 @@
 ### Release notes
 
-- Cached IModel instead of opening one per connection.
+- Cached IModel instead of opening one per connection. [#148](https://github.com/Azure/azure-functions-rabbitmq-extension/issues/148)
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,1 +1,3 @@
 ### Release notes
+
+- Cached IModel instead of opening one per connection.
--- a/src/Bindings/RabbitMQClientBuilder.cs
+++ b/src/Bindings/RabbitMQClientBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using Microsoft.Extensions.Options;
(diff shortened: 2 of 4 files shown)


release_notes.md
```diff
@@ -1 +1,3 @@
 ### Release notes
+
+- Cached IModel instead of opening one per connection.
```
pragnagopa: nit: Link to issue

pragnagopa (approved):

PR #164, opened 2021-05-10
Add parameters to set Acceptable policy errors when connecting over TLS

kzhen:
When making a connection to the RabbitMQ broker over TLS the RabbitMQ .Net Client rejects the connection if the certificates configured on the broker are self signed and not considered 'trusted'. This PR allows the passing of the `SslPolicyErrors` enum as a parameter to indicate which errors are acceptable for TLS connections.

### DIFF
--- a/src/Services/RabbitMQService.cs
+++ b/src/Services/RabbitMQService.cs
@@ -2,6 +2,9 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Net.Security;
+using System.Reflection;
+using System.Security.Cryptography.X509Certificates;
 using RabbitMQ.Client;
 
 namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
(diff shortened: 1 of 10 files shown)


src/Services/RabbitMQService.cs
```diff
@@ -17,22 +20,24 @@ internal sealed class RabbitMQService : IRabbitMQService
...
             _userName = userName;
             _password = password;
             _port = port;
+            _acceptablePolicyErrors = acceptablePolicyErrors;
 
-            ConnectionFactory connectionFactory = GetConnectionFactory(_connectionString, _hostName, _userName, _password, _port);
+            ConnectionFactory connectionFactory = GetConnectionFactory(_connectionString, _hostName, _userName, _password, _port, _acceptablePolicyErrors);
 
-            _model = connectionFactory.CreateConnection().CreateModel();
+            _model = connectionFactory.CreateConnection(Assembly.GetEntryAssembly().GetName().Name).CreateModel();
```
kzhen, line 36: When creating the connection we pass through the Assembly name so the connection can be identified in the RabbitMQ management portal

kzhen:
@yojagad any thoughts on this?

JatinSanghvi:
Thanks for the PR. I found that #178 is trying to accomplish a similar goal. I personally like the idea of giving user the ability to define the acceptable SSL errors as being done in this PR, but there should be also a way to enable/disable SSL if user is passing binding parameters individually (instead of connection string with scheme `amqp` or `amqps`) as it is being done in the other PR. Since both changes are conflicting, I will need to close this one in favor of the other one, as that one is the latest one and I have already started following up on it actively. I will also reconsider on whether connection name should be passed on not.

kzhen:
Thanks for the update @JatinSanghvi, the original intention that I had with this PR was to use this NuGet package for some work related projects. Instead we've gone in the direction of writing our own custom trigger bindings which wrap EasyNetQ - that has the added advantage of taking care of an opinionated approach to queue/exchange creation as well as bindings between queues and exchanges.

The code in the repository helped a lot to get an understanding of the different parts involved in writing such a library - so thanks to all the team 😃

PR #167, opened 2021-08-24
Service pooling with rabbitmq

yojagad:
RabbitMQ service is not being pooled for every output binding invocation. As a result, every invocation of IAsyncCollector creates a new service.

This PR pools instances of IRabbitMQService. This also means that Batch needs to be reset upon every `FlushAsync()` call and the PR handles that as well.

### DIFF
--- a/src/Config/RabbitMQExtensionConfigProvider.cs
+++ b/src/Config/RabbitMQExtensionConfigProvider.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
         private readonly IRabbitMQServiceFactory _rabbitMQServiceFactory;
         private readonly ILogger _logger;
         private readonly IConfiguration _configuration;
-        private readonly ConcurrentDictionary<string, IRabbitMQService> _connectionParamtersToService;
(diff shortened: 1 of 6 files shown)


src/Config/RabbitMQExtensionConfigProvider.cs
```diff
@@ -108,13 +111,15 @@ internal RabbitMQContext CreateContext(RabbitMQAttribute attribute)
 
         internal IRabbitMQService GetService(string connectionString, string hostName, string queueName, string userName, string password, int port)
         {
-            return _rabbitMQServiceFactory.CreateService(connectionString, hostName, queueName, userName, password, port);
+            string key = string.Join(connectionString, ",", hostName, ",", queueName, ",", userName, ",", password, ",", port);
```
TsuyoshiUshio: Nit You can reduce the code by this approach. 

```
            var keyArray= { connectionString, hostName .... };
            string key = String.Join(",", keyArray);
```

TsuyoshiUshio (approved):
Overall design looks good. I just leave small comment. However, the change will be up to you.
If you have E2E testing it might be great for this change.

PR #170, opened 2021-08-25
Add lock for RabbitMQService

yojagad:
Added lock to ensure `batch.Add()` and `flush()` don't happen simultaneously.

### DIFF
--- a/src/WebJobs.Extensions.RabbitMQ.csproj
+++ b/src/WebJobs.Extensions.RabbitMQ.csproj
@@ -9,7 +9,7 @@
     <SignAssembly>false</SignAssembly>
     <AssemblyOriginatorKeyFile>PublicKey.snk</AssemblyOriginatorKeyFile>
     <DelaySign>true</DelaySign>
-    <Version>1.2.0</Version>
+    <Version>1.1.0</Version>
     <PackageId>Microsoft.Azure.WebJobs.Extensions.RabbitMQ</PackageId>
     <Authors>Microsoft</Authors>
     <Product>Azure WebJobs RabbitMQ Extension</Product>
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,1 +1,3 @@
 ### Release notes
+
+# Added pooling for RabbitMQService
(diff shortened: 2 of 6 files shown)


src/WebJobs.Extensions.RabbitMQ.csproj
```diff
@@ -9,7 +9,7 @@
     <SignAssembly>false</SignAssembly>
     <AssemblyOriginatorKeyFile>PublicKey.snk</AssemblyOriginatorKeyFile>
     <DelaySign>true</DelaySign>
-    <Version>1.2.0</Version>
+    <Version>1.1.0</Version>
```
TsuyoshiUshio, line 12: Just curious why version down?

TsuyoshiUshio (approved):
Generally speaking I approve. It's up to you. 
However, I thought, the lock scope looks big. I mean if the IModel should be protected, then we can scope down it to just inside of the  https://github.com/Azure/azure-functions-rabbitmq-extension/blob/0b1358c4618b207cec89377fa83cacfdb7fe27a6/src/Services/RabbitMQService.cs Then we might be shorten the time of the locking.  I might too nit picking. So that Up to you. :)

src/WebJobs.Extensions.RabbitMQ.csproj
yojagad: I cut a release and then realized this. Hence version down..

JatinSanghvi:
Hi @yojagad, was this one a proactive change or a fix to a known issue? Was it the case that if multiple function invocations were running in parallel, there was possibility that `Add` on shared `BasicPublishBatch` instance could be called between `Publish` and `ResetPublishBatch` and that could have resulted in lost messages?

yojagad:
> Hi @yojagad, was this one a proactive change or a fix to a known issue? Was it the case that if multiple function invocations were running in parallel, there was possibility that `Add` on shared `BasicPublishBatch` instance could be called between `Publish` and `ResetPublishBatch` and that could have resulted in lost messages?

This was a proactive follow up to service pooling PR here https://github.com/Azure/azure-functions-rabbitmq-extension/pull/167 . This wasn't an issue when connections (for output binding) weren't pooled.

PR #95, opened 2020-10-15
Add new parameter to specify the vhost to use on RabbitMQ server.

Socolin:
See https://www.rabbitmq.com/vhosts.html for more details.
It was already possible to use this using the ConnectionString, this allow it
to be used when specifying host/user/password

Fixes: #81

Issue #81, opened 2019-11-22
Add support for connecting to user defined Virtual Hosts


markm9199:
If I've understood correctly then this will only connect to queues in the Virtual Host called \

I had to give my user account access to \ and move the queue I was testing with into \ otherwise it would not work.

Are there any plans to extend to user defined Virtual Hosts?

Also, any plans to be able to publish to an Exchange with a routing key rather than directly to a queue?

RobBowman:
Any update on this please?

ThejaChoudary:
@markm9199  Apologies for the late reply 
Can you please provide us with more info about this to investigate further and also please refer the [link](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/dedicated-hosts)  for your guidance.


Thank You for your patience .

markm9199:
I don't understand why you have asked me to look at a link to Azure Dedicated Hosts, I think perhaps there is a misunderstanding? 

I am talking about a feature of RabbitMQ called Virtual Hosts. It is normal in a large Rabbit MQ installation, to group exchanges and queues using different Virtual Hosts, on the Rabbit MQ. I assume as a rabbit extension developer you are familiar with this concept?

My issue is that this extension does not allow me to specify which Virtual Host to use, in fact it defaults to the top level Virtual Host , called '\\'. **I would like to be able to specify the Rabbit Virtual Host**. 

My business has many exchanges and queues on a single Rabbit MQ cluster, and they are grouped into Virtual Hosts. I cannot change this and so if I cannot specify the Virtual Host then I cannot connect to the my queues with this extension.

Here is an image snipped from a Rabbit MQ management console, showing a section where one can select and create Virtual Hosts. This is just to indicate to you what I mean by Virtual Hosts.

![image](https://user-images.githubusercontent.com/51416865/88288370-9953ac80-cceb-11ea-853c-846b79386e65.png)

Is this enough information?

markm9199:
@ThejaChoudary any update after my recent comments?

ThejaChoudary:
@pragnagopa --Can you please take a look .Thank you !

pragnagopa:
@markm9199  - Thanks for the feature request.

Moved this to Triaged.

Socolin:
This seems to be already possible to use vhosts using connection string like
```
  "connectionStrings": {
    "rabbitMQ": "amqp://localhost/azure-rabbitmq"
  }
```
[...]

PR #171, opened 2021-08-27
Add support for VHost in connection string.

kentonbmax:
Add support for parsing vhost from connection property and supplying it to the rabbit client.

### DIFF
--- a/src/Services/RabbitMQService.cs
+++ b/src/Services/RabbitMQService.cs
@@ -90,6 +90,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
                 }
             }
 
+            var rabbitUri = new Uri(connectionString);
+
+            if (rabbitUri.AbsolutePath != "/")
+            {
+                connectionFactory.Uri = new Uri(rabbitUri.AbsoluteUri.Replace(rabbitUri.AbsolutePath, default));
+                connectionFactory.VirtualHost = rabbitUri.AbsolutePath;
+            }
+
             return connectionFactory;
         }
     }
--- a/src/Services/RabbitMQService.cs
+++ b/src/Services/RabbitMQService.cs
@@ -90,12 +90,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
                 }
             }
 
-            var rabbitUri = new Uri(connectionString);
-
-            if (rabbitUri.AbsolutePath != "/")
+            if (!string.IsNullOrWhiteSpace(connectionString))
             {
-                connectionFactory.Uri = new Uri(rabbitUri.AbsoluteUri.Replace(rabbitUri.AbsolutePath, default));
-                connectionFactory.VirtualHost = rabbitUri.AbsolutePath;
+                var rabbitUri = new Uri(connectionString);
+
+                if (rabbitUri.AbsolutePath != "/")
+                {
+                    connectionFactory.Uri = new Uri(rabbitUri.AbsoluteUri.Replace(rabbitUri.AbsolutePath, default));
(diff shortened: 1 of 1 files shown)


kentonbmax:
Fix for #81

kentonbmax:
@pragnagopa  is this repo still active? There is not a lot of activity here yet this is still in use?

pragnagopa:
> @pragnagopa is this repo still active? There is not a lot of activity here yet this is still in use?

Apologies for late response here. Yes, this repo is still active. Tagging @JatinSanghvi @lpapudippu to follow up

src/Services/RabbitMQService.cs
```diff
@@ -90,6 +90,17 @@ internal static ConnectionFactory GetConnectionFactory(string connectionString,
                 }
             }
 
+            if (!string.IsNullOrWhiteSpace(connectionString))
```
JatinSanghvi, line 93: I have a couple of asks:
1. Why not merge the `if (rabbitUri.AbsolutePath != "/")` with line 68?
2. Line 99 is not safe as the hostname might contain the VHost string. Use `connectionFactory.Uri = new Uri(rabbitUri.GetLeftPart(UriPartial.Authority));` instead.
3. I am new to Rabbit MQ. The connection string used in code example on https://www.cloudamqp.com/docs/dotnet.html#publish would set the Virtual Host to `%2f` but either the VHost is not present here or its value is `/` (need to check which one). I found the connection string examples in **Appendix A: Examples** on https://www.rabbitmq.com/uri-spec.html. A correct fix here would be to URI decode the path and then set the VHost and other properties.
4. Please update the test case in file `test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQServiceTests.cs` to ensure that expected value of VHost string is assigned to `connectionFactory.VirtualHost`.

I also think that we should update the `RabbitMQService` constructor to explicitly consume VHost as an argument. May need to ensure that it does not cause a breaking change. Anyhow that can be done as a separate step.

JatinSanghvi (changes requested):

src/Services/RabbitMQService.cs
JatinSanghvi: Hi @kentonbmax, I created a test app with code as below. I did not require to manually extract VHost from the connection string. It was set automatically by `ConnectionFactory` on setting its `Uri` property. I think I am using same RabbitMQ client library version as this extension, but am on .NET Core 3.1 instead of 2.2. Let me know if I am missing something here.

**Program**

```cs
// CloudAMQP URL in format amqp://user:pass@hostName:port/vhost
var url = "amqp://user%61:%61pass@host:10000/v%2fhost";

var factory = new ConnectionFactory
{
    Uri = new Uri(url)
};

Console.WriteLine(factory.UserName);
Console.WriteLine(factory.Password);
Console.WriteLine(factory.HostName);
Console.WriteLine(factory.Port);
Console.WriteLine(factory.VirtualHost);
```

**Output**
```text
usera
apass
host
10000
v/host
```

kentonbmax: I did try my connection string with /env in my existing azure functions where the connection string is an environment variable and the trigger failed to connect. I believe the issue is in the constructor resetting the values. So my question is why explicitly set anything other then connectionString?

JatinSanghvi: As mentioned in https://github.com/Azure/azure-functions-rabbitmq-extension/issues/81#issuecomment-708888734, connection-string is the workaround for lack of support for virtual host parameter. I think the extension supports both combinations of binding parameters (connection string or passing each piece of information individually) and we will be keeping it like this. I will need to close this PR and reopen #95.

JatinSanghvi:
Closing this one in favor of #95.

PR #174, opened 2021-10-20
Update RabbitMQ client version to 6.2.2

### LINKED ISSUE #92: Support for RabbitMQ client 6.+
Currently the extensions does not work with the Rabbit MQ client versions 6.+

Mainly it needs to be changed to support the change in body, from using byte arrays to ReadOnlyMemory.

### LINKED ISSUE #173: Message properties are not being sent in GRPC payload to out of process worker due to bug in RabbitMQ.Client 5.1.2
When a producer sends a message with message properties, it is not being sent to the out-of-process worker, instead the type name string is being sent.

**Version:** WebJobs.Extensions.RabbitMQ version 1.1.0

_Note: This is not a code bug in the rabbitmq-extension, but a bug in the RabbitMQ.Client version 5.1.2, which we are using for WebJobs.Extensions.RabbitMQ 1.1.0_

#### Repro steps
1. Create a dotnet isolated worker to be used as a consumer of the message.
2. Publish a message with any properties other than `reply_to`. You may use the RabbitMQ admin portal to do this.
3. In the dotnet consumer project, setup a break point. Inspect the binding data on function context for `BasicProperties`. You can see that the value is a string of the type name (`RabbitMQ.Client.Framing.BasicProperties`) instead of a complex object. 

#### Why is this happening ?

I did a bit of investigation on why this is happening. Apparently, the host is trying to serialize the `BasicProperties` object while trying to create the GRPC payload to send to the worker, but it throws because accessing the `ReplyToAddress` property on `BasicProperties` object is throwing an exception. When an exception happens while trying to serialize the object, we call the `ToString` on the object in the catch block. In our case, the `ToString` call on this type is producing the type name value string.

rabbitmq-host-throwing2

The `WebJobs.Extensions.RabbitMQ version 1.1.0` is using `RabbitMQ.Client 5.1.2` package. If you look at the `ReplyToAddress` property implementation on 5.1.2, it is returning the result of `PublicationAddress.Parse(this.ReplyTo)`. 

rabbitmqclient-replytoaddress-implementation

The `PublicationAddress.Parse` method is internally calling the `Regex.Match` method and passing the `reply_to` property value. [This method](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.match?view=net-5.0#System_Text_RegularExpressions_Regex_Match_System_String_) will throw an `ArgumentNullException` when the input is `null`, which is what is happening in our case.

rabbitmqclient-parse-method

#### Mitigation

We will advise the users to always specify the `reply_to` property when publishing a message.

#### Fix

The RabbitMQ team had [fixed](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/828) this issue in May 2020. The fix is published in their [6.1.0](https://www.nuget.org/packages/RabbitMQ.Client/6.1.0) release. We should publish a new version of `WebJobs.Extensions.RabbitMQ` consuming the 6.1.0 RabbitMQ.Client package.

JatinSanghvi:
- Updated RabbitMQ client version from `5.1.2` to `6.2.2`.
- Updated the target framework for sample and test projects. '.NET Core 2.2' is out of support since 2019.
- Updated the code to work with a [breaking change](https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/af59ca90fefd7e473cc235f711881984e892d7ea#diff-43a441e373cb8dac4813a43ff4fcf12165a477b2e41944aab61d3d98ce96b841) introduced in `6.2.2` client.
- Fixed unit test to work with new client version.
- Verified all sample functions work as expected with the changes.

Fixes #173, #92


src/WebJobs.Extensions.RabbitMQ.csproj
```diff
@@ -47,7 +47,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.14" />
-    <PackageReference Include="RabbitMQ.Client" Version="5.1.2" />
+    <PackageReference Include="RabbitMQ.Client" Version="6.2.2" />
```
kshyju, line 49: This fix should be released in a new version of [Microsoft.Azure.WebJobs.Extensions.RabbitMQ](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.RabbitMQ/) package. Looks like we are missing that version bump change in this PR.

kshyju (dismissed):

src/WebJobs.Extensions.RabbitMQ.csproj
JatinSanghvi: Right, I need to make the build succeed, bump up the version and run the samples as validation before requesting for PR review. I have marked the PR as draft to correctly reflect its status. I would try to make the PR ready for review in early next week.

azure-pipelines.yml
```diff
@@ -59,7 +59,7 @@ jobs:
   - task: UseDotNet@2
     inputs:
       packageType: 'sdk'
-      version: '2.x'
+      version: '6.x'
```
JatinSanghvi, line 62: Updated target framework to .NET 6.0 for sample and test projects. .NET Core 2.2 is out of support since two years.

src/Trigger/IRabbitMQModel.cs
```diff
@@ -26,7 +27,7 @@ public interface IRabbitMQModel
 
         void BasicReject(ulong deliveryTag, bool requeue);
 
-        void BasicPublish(string exchange, string routingKey, IBasicProperties basicProperties, byte[] body);
+        void BasicPublish(string exchange, string routingKey, IBasicProperties basicProperties, ReadOnlyMemory<byte> body);
```
kshyju, line 30: [This interface is public](https://github.com/Azure/azure-functions-rabbitmq-extension/blob/dev/src/Trigger/IRabbitMQModel.cs#L9). So, changing the method signature will be a breaking change for anyone using this API. Do you know who uses this?

Do we need this change? If so, should we do it in a backward compatible way? (Introduce a new overload and mark the old one as obsolete).

kshyju (commented):
One question about the public API surface.

yojagad:
The general guidance is to introduce newer rabbitmq client version updates (major version updates) through a new major version of the extension.

src/Trigger/IRabbitMQModel.cs
JatinSanghvi: There's no public class that accepts `IRabbitMQModel` as its method argument or has `IRabbitMQModel` as return type. The only path where I could find the model definition leaking was via `IRabbitMQModel` <- `IRabbitMQService` <-`IRabbitMQServiceFactory` but it stops at that point. There are no public classes taking or returning either `IRabbitMQService` or `IRabbitMQServiceFactory` or any of their implementation classes. I guess customer would not be impacted if they cannot work with `IRabbitMQModel` in connection with any other classes in RabbitMQ extension. I will fix the visibility of the classes and interfaces in a future PR.

JatinSanghvi:
> The general guidance is to introduce newer rabbitmq client version updates (major version updates) through a new major version of the extension.

Following the same tradition as of other Azure Functions repo, I will fork out `v1.x` branch but `dev` branch will continue to be the one for next minor/major release which will be `2.0.0`. So the PR has correct branch set as the target.

lpapudippu (approved):

PR #176, opened 2021-12-16
Prevent package reference RabbitMQ.Client v 6.x

### LINKED ISSUE #89: Method not found: 'Byte[] RabbitMQ.Client.Events.BasicDeliverEventArgs.get_Body()'.
Hi,

I have hosted RabbitMQ in a Docker container and via a simple Asp.Net Core web site confirmed that I can send messages. The management console confirms.

I have created a simple Azure Function, run locally, but when it connects, it gets the error when the function runs in debug on my dev box:
     
    Method not found: 'Byte[] RabbitMQ.Client.Events.BasicDeliverEventArgs.get_Body()'.

I do get a consumer listed once the Function runs, but the Function crashes after it gets the above error.

image

My function, run locally in Visual Studio 2019, is very simple:

     public static class TriggerFunction
     {
        [FunctionName("TriggerFunction")]
        public static void Run(
            [RabbitMQTrigger("hello")] string message, ILogger log)
        {
            log.LogInformation($"************* Message received from RabbitMQ trigger: {message}");
        }
     }

My NuGets are:

     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.RabbitMQ" Version="0.2.2029-beta" />
     <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.7" />
     <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.10.9" />
     <PackageReference Include="RabbitMQ.Client" Version="6.0.0" />

Cogax:
As we don't have support for `RabbitMQ.Client` 6.x #92, we should prevent that version. This fixes #89.

### DIFF
--- a/src/WebJobs.Extensions.RabbitMQ.csproj
+++ b/src/WebJobs.Extensions.RabbitMQ.csproj
@@ -47,7 +47,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.14" />
-    <PackageReference Include="RabbitMQ.Client" Version="5.1.2" />
+    <PackageReference Include="RabbitMQ.Client" Version="[5.1.2,6.0)" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.118">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>


JatinSanghvi:
Thanks for sending the PR. We have upgraded RabbitMQ client version to 6.2.2 (#174) so this PR won't be needed. We should get 2.0.0 pre-release version of extension out in early February.

PR #178, opened 2022-01-13
Allow SSL for AMQP connection

aaguilartablada:
In order to connect to an AMQPS endpoint a new attribute is necessary to configure 'SslOption' for 'ConnectionFactory'.
Furthermore, to allow using self-signed certificates for development environments another attribute has been introduced: 'InsecureSsl'.
By default, SslOption does NOT enable SNI extension (https://en.wikipedia.org/wiki/Server_Name_Indication). To enable it, setting the property 'ServerName' is enough. This extension is mandatory to allow connecting to several RabbitMQ clusters behind a load balancer.

### DIFF
--- a/src/Services/RabbitMQService.cs
+++ b/src/Services/RabbitMQService.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Net.Security;
 using RabbitMQ.Client;
 
 namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
@@ -17,11 +18,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
         private readonly string _password;
         private readonly int _port;
         private readonly bool _ssl;
+        private readonly bool _insecureSsl;
         private readonly object _publishBatchLock;
 
         private IBasicPublishBatch _batch;
 
-        public RabbitMQService(string connectionString, string hostName, string userName, string password, int port, bool ssl)
+        public RabbitMQService(string connectionString, string hostName, string userName, string password, int port, bool ssl, bool insecureSsl)
         {
             _connectionString = connectionString;
             _hostName = hostName;
@@ -29,15 +31,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
             _password = password;
             _port = port;
             _ssl = ssl;
+            _insecureSsl = insecureSsl;
 
-            ConnectionFactory connectionFactory = GetConnectionFactory(_connectionString, _hostName, _userName, _password, _port, _ssl);
+            ConnectionFactory connectionFactory = GetConnectionFactory(_connectionString, _hostName, _userName, _password, _port, _ssl, _insecureSsl);
 
             _model = connectionFactory.CreateConnection().CreateModel();
             _publishBatchLock = new object();
         }
 
-        public RabbitMQService(string connectionString, string hostName, string queueName, string userName, string password, int port, bool ssl)
-            : this(connectionString, hostName, userName, password, port, ssl)
+        public RabbitMQService(string connectionString, string hostName, string queueName, string userName, string password, int port, bool ssl, bool insecureSsl)
+            : this(connectionString, hostName, userName, password, port, ssl, insecureSsl)
         {
             _rabbitMQModel = new RabbitMQModel(_model);
             _queueName = queueName ?? throw new ArgumentNullException(nameof(queueName));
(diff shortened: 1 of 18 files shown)


ghost (approved):

aaguilartablada:
@yojagad @JatinSanghvi Hello! Could you take a look at this PR, please? Thank you so much.

src/Utility.cs
```diff
@@ -28,6 +28,19 @@ internal static int FirstOrDefault(params int[] values)
             });
         }
 
+        internal static bool FirstOrDefault(params bool[] values)
+        {
+            return values.FirstOrDefault(v =>
```
JatinSanghvi: Can you update to `return values.FirstOrDefault(v => v);`  and similarly the above method body to `return values.FirstOrDefault(v => v != 0);`?

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -60,5 +60,15 @@ public RabbitMQTriggerAttribute(string hostName, string userNameSetting, string
...
         /// </summary>
         public int Port { get; set; }
+
+        /// <summary>
+        /// Enable or disable ssl in RabbitMQ connection.
+        /// </summary>
+        public bool Ssl { get; set; }
+
+        /// <summary>
+        /// Enable os disable checking certificate when Ssl is enabled (not recommended for production).
```
JatinSanghvi, line 70: It would be better to have more descriptive attribute properties like `EnableSsl` and `SkipCertificateValidation`.

src/Services/RabbitMQService.cs
```diff
@@ -88,9 +95,30 @@ internal static ConnectionFactory GetConnectionFactory(string connectionString,
...
+        {
+            if (ssl)
+            {
+                connectionFactory.Ssl = new SslOption
+                {
+                    Enabled = true,
+
+                    // Set SNI in order to work for multiple RabbitMQ clusters located behind a LoadBalancer
+                    ServerName = hostname,
+                };
```
JatinSanghvi, line 115: Question: If certificate path is not supplied, wouldn't there be `RemoteCertificateNotAvailable` error?

JatinSanghvi (changes requested):

src/Trigger/RabbitMQTriggerAttribute.cs
JatinSanghvi: Found #164 which is accomplishing same stuff. Can you check that one and see if `bool insecureSsl` can be replaced with `SslPolicyErrors acceptablePolicyErrors` so user has freedom of which errors to allow?

JatinSanghvi:
This will also require updating [RabbitMQOutput.java](https://github.com/Azure/azure-functions-rabbitmq-extension/blob/dev/binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/RabbitMQOutput.java), [RabbitMQTrigger.java](https://github.com/Azure/azure-functions-rabbitmq-extension/blame/dev/binding-library/java/src/main/java/com/microsoft/azure/functions/rabbitmq/annotation/abbitMQTrigger.java) and respective test cases.

src/Services/RabbitMQService.cs
aaguilartablada: Presenting the certificate without the whole path is not a problem if you set RemoteCertificateChainErrors as an acceptable error. 
Not presenting any certificate will always produce a RabbitMQ.Client.Exceptions.BrokerUnreachableException exception.

src/Trigger/RabbitMQTriggerAttribute.cs
aaguilartablada: According to https://docs.microsoft.com/es-es/dotnet/api/system.net.security.sslpolicyerrors , the interesting errors when skipping certificate validation are RemoteCertificateChainErrors and RemoteCertificateNameMismatch and only for development or testing purposes. So, in my opinion they have to be together.

The RemoteCertificateNotAvailable never should be accepted, as it will always produce an exception when trying to establish a SSL session.

For these reasons, in my opinion, it is better setting the attribute/option as a boolean to simplify code. Those who activate SkipCertificateValidation have to know what they are doing and do it under their own risk.

Do you agree?

src/Utility.cs
aaguilartablada: sure!

aaguilartablada:
Changes requested done and Java files updated. Thanks for your review @JatinSanghvi

JatinSanghvi (approved):

JatinSanghvi:
Thanks @aaguilartablada, I have approved the pull request. If possible, I would like to know how frequently you need to supply individual tokens (hostname, username, password, port, ssl) instead of just the connection string. Most of other extensions are allowing just connection string to be passed.

aaguilartablada:
Thaks so much @JatinSanghvi. I use connection string, but I applied the changes in this PR also for individual tokens in case someone using them need to configure SSL this way.

When do you plan to release a new version including these changes? Thanks.

JatinSanghvi:
I am planning to get 2.0.0-preview1 released this week if everything goes well. The release will unblock users who are encountering issue #173. This would be a pre-release as there are more breaking changes planned for future releases. I am thinking of only retaining connection string since the attribute properties and internal method signatures are getting complex with introduction of new connection-string elements (next one in sequence was going to be a `vhost` property).

aaguilartablada:
Perfect! Thanks again. Good luck ;)

PR #180, opened 2022-01-29
Add new parameter to specify the vhost to use on RabbitMQ server.

### LINKED ISSUE #81: Add support for connecting to user defined Virtual Hosts

Socolin:
See https://www.rabbitmq.com/vhosts.html for more details.
It was already possible to use this using the ConnectionString, this allow it
to be used when specifying host/user/password

Fixes: #81

### DIFF
--- a/src/RabbitMQAttribute.cs
+++ b/src/RabbitMQAttribute.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
@@ -46,6 +46,13 @@ namespace Microsoft.Azure.WebJobs
         /// Gets or sets the Port used. Defaults to 0.
         /// </summary>
         public int Port { get; set; }
+
+        /// <summary>
+        /// Gets or sets the vhost to use when connecting to RabbitMQ using HostName / UserName / Password
+        /// <see href="https://www.rabbitmq.com/vhosts.html"/>.
+        /// </summary>
+        [AppSetting]
+        public string VirtualHost { get; set; }
 
         /// <summary>
         /// Gets or sets the name of app setting that contains the connection string to authenticate with RabbitMQ.
--- a/test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQOptionsTests.cs
+++ b/test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQOptionsTests.cs
@@ -21,6 +21,7 @@ namespace WebJobs.Extensions.RabbitMQ.Tests
                 { nameof(option.QueueName), option.QueueName },
                 { nameof(option.Port), option.Port },
                 { nameof(option.PrefetchCount), option.PrefetchCount },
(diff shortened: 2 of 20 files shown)


Socolin:
I've updated the code with the latest version since previous PR was closed 
https://github.com/Azure/azure-functions-rabbitmq-extension/pull/95#issuecomment-1024896888

I did no manual tests with the updated code. Unit tests are passing.

JatinSanghvi:
> I've updated the code with the latest version since previous PR was closed [#95 (comment)](https://github.com/Azure/azure-functions-rabbitmq-extension/pull/95#issuecomment-1024896888)
> 
> I did no manual tests with the updated code. Unit tests are passing.

Thanks, I will get it manually tested.

src/RabbitMQAttribute.cs
```diff
@@ -47,6 +47,13 @@ public sealed class RabbitMQAttribute : Attribute
         /// </summary>
         public int Port { get; set; }
 
+        /// <summary>
+        /// Gets or sets the vhost to use when connecting to RabbitMQ using HostName / UserName / Password
+        /// <see href="https://www.rabbitmq.com/vhosts.html"/>.
+        /// </summary>
+        [AppSetting]
```
JatinSanghvi, line 54: Delete line 54. `AppSetting` attribute will require the value for `VirtualHost` parameter to always come from settings file. We add that restriction only for parameters who values might be secrets.

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQOptionsTests.cs
```diff
@@ -21,6 +21,7 @@ private string GetFormattedOption(RabbitMQOptions option)
                 { nameof(option.QueueName), option.QueueName },
                 { nameof(option.Port), option.Port },
                 { nameof(option.PrefetchCount), option.PrefetchCount },
+                { nameof(option.VirtualHost), option.VirtualHost },
```
JatinSanghvi, line 24: Swapping lines 23 and 24 should make the unit tests run successfully.

JatinSanghvi (changes requested):

JatinSanghvi:
@Socolin, I have done manual E2E validation of the changes. If you can make the requested changes, I will approved and merge the PR.

JatinSanghvi:
@Socolin, we just introduced `enableSsl` parameter to both trigger and output bindings. With this PR, there will be yet another parameter introduced. In general, based on my discussion with other developers, most of them use connection strings instead of passing individual tokens (username, password, hostname, port, etc.) in their bindings. The other official extensions only have support for connection strings though something like (account-name + account-key) was possible for them. Hence, I have decided to remove the attribute-properties for individual parts. The reason why we need to introduce `vhost` parameter in this PR is to overcome the limitation of only working with default `vhost` when user passes username, password, etc. individually. Supporting only the connection string should avoid the need of this parameter. Considering this, I have decided to close the PR. Let me know if you have any comments.. thanks for the PR though.

PR #184, opened 2022-02-17
Remove connection-string elements from attributes

JatinSanghvi:
Current extension supports both RabbitMQ connection URI e.g. `amqp://user:pass@host:10000/vhost` and individual connection-string elements (`user`, `pass`, `host`, and `1000`) to be specified in trigger and output bindings. With the latter way to specify connection information, some of the elements were still missing, namely isSslEnabled and vHost. We recently added commit to include attribute property for isSslEnabled. Instead of adding and maintaining these many number of attribute properties, this commit removes the latter support. This would also eliminate combinations like hostname coming from appsettings but port coming from attributes, which is not what we expect from users.

As per my research, most of the users use connection strings in their functions. Since next one would be a version 2.x release, I am taking the opportunity to introduce this breaking change. Users using connection strings already are not expected to update their functions to upgrade to v2.x release. The other WebJobs extensions from Azure only support connection strings, as far as I have seen.


.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+    "recommendations": [
+        "ms-dotnettools.csharp"
+    ]
+}
```
JatinSanghvi, line 5: Added a few files to make it easier to code on Visual Studio Code. Users would normally open the root folder to work on C# code, and `/java-library` folder for Java code.

azure-pipelines.yml
```diff
@@ -167,19 +167,19 @@ jobs:
...
         inputs:
-          mavenPomFile: binding-library\java\pom.xml
+          mavenPomFile: java-library\pom.xml
           options: --batch-mode --define=gpg.skip --define=org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn --update-snapshots
 
       - powershell: |
           $prefix = 'azure-functions-java-library-rabbitmq-$(version)'
-          $source = 'binding-library\java'
+          $source = 'java-library'
           $destination = '$(Build.ArtifactStagingDirectory)\$(version)'
```
JatinSanghvi, line 183: Updated folder name for Java code to match more with https://github.com/Azure/azure-functions-java-library/ where Java annotations for most of other extensions reside. Also made minor change to build job names. The job names show up on every PR page in PR validation section.

src/Bindings/RabbitMQClientBuilder.cs
```diff
@@ -31,14 +31,9 @@ private IModel CreateModelFromAttribute(RabbitMQAttribute attribute)
...
-            string resolvedUserName = Utility.FirstOrDefault(attribute.UserName, _options.Value.UserName);
-            string resolvedPassword = Utility.FirstOrDefault(attribute.Password, _options.Value.Password);
-            int resolvedPort = Utility.FirstOrDefault(attribute.Port, _options.Value.Port);
-            bool resolvedEnableSsl = Utility.FirstOrDefault(attribute.EnableSsl, _options.Value.EnableSsl);
-            bool resolvedSkipCertificateValidation = Utility.FirstOrDefault(attribute.SkipCertificateValidation, _options.Value.SkipCertificateValidation);
-
-            IRabbitMQService service = _configProvider.GetService(resolvedConnectionString, resolvedHostName, resolvedUserName, resolvedPassword, resolvedPort, resolvedEnableSsl, resolvedSkipCertificateValidation);
+            bool resolvedDisableCertificateValidation = Utility.FirstOrDefault(attribute.DisableCertificateValidation, _options.Value.DisableCertificateValidation);
+
+            IRabbitMQService service = _configProvider.GetService(resolvedConnectionString, resolvedDisableCertificateValidation);
```
JatinSanghvi, line 37: Removed connection string elements: `HostName`, `UserName`, `Password`, `Port`, and `EnableSsl` from both trigger and output binding attributes. Renamed `SkipCertificateValidation` to `DisableCertificateValidation`. The `SkipCertificateValidation` property was added recently and was not released to users.

src/Services/RabbitMQService.cs
```diff
@@ -12,35 +12,33 @@ internal sealed class RabbitMQService : IRabbitMQService
...
-            ConnectionFactory connectionFactory = GetConnectionFactory(_connectionString, _hostName, _userName, _password, _port, _enableSsl, _skipCertificateValidation);
+            ConnectionFactory connectionFactory = new ConnectionFactory
+            {
+                Uri = new Uri(connectionString),
+            };
+
+            if (disableCertificateValidation && connectionFactory.Ssl.Enabled)
+            {
+                connectionFactory.Ssl.AcceptablePolicyErrors |= SslPolicyErrors.RemoteCertificateChainErrors;
+            }
```
JatinSanghvi, line 35: Method `GetConnectionFactory` is not required. The `connectionFactory.Ssl` (line 31 and 33) will be set to non-null value by RabbitMQ client after we set its `Uri` property (line 28).

test/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQServiceTests.cs
```diff
@@ -1,41 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
```
JatinSanghvi, line 1: The single test case is no longer required, hence the file was deleted.

binding-library/java/release_notes.md
```diff
@@ -1 +0,0 @@
-### Release notes
```
JatinSanghvi, line 1: The release notes are empty at the time of PR. The intension here was to add all changes that will go in next minor version in the notes, and then move its content in GitHub release notes when a new extension is released, and empty up the files again so it's ready for next version notes. I have updated the release pipelines already that does not treat release notes files specially.

I would be replacing both files with a single file containing change-log across all versions. Adding notes for GitHub release manually allows for more customizations, such as thanking GitHub contributors, giving an overview on release, details on breaking changes, etc.

src/Services/RabbitMQService.cs
```diff
@@ -62,63 +60,5 @@ public void ResetPublishBatch()
...
-        internal static void ConfigureSsl(ConnectionFactory connectionFactory, string hostname, bool enableSsl, bool skipCertificateValidation)
-        {
-            if (enableSsl)
-            {
-                connectionFactory.Ssl = new SslOption
-                {
-                    Enabled = true,
-
-                    // Set SNI in order to work for multiple RabbitMQ clusters located behind a LoadBalancer
-                    ServerName = hostname,
```
JatinSanghvi, line 114: Received above code through contribution a few days back. With this PR, the `ConnectionFactory` will always be initialized with RabbitMQ connection URI. The RabbitMQ client [sets the server name](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/389819a82b9c77811da96db586c811853ec99b14/projects/RabbitMQ.Client/client/api/ConnectionFactory.cs#L567) in this case, so we do not need to set it by ourselves.

src/Trigger/RabbitMQTriggerAttribute.cs
```diff
@@ -6,69 +6,37 @@
 
 namespace Microsoft.Azure.WebJobs
 {
+    /// <summary>
+    /// Attribute used to bind a parameter to RabbitMQ trigger message.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
```
JatinSanghvi, line 12: Constraining the attribute to be applied to function parameters only.

v-shenoy (approved):
Nice work, @JatinSanghvi.
The changes look thorough.

PR #185, opened 2022-02-21
Update namespaces and assembly names

JatinSanghvi:
- Updated namespaces and assembly names to be consistent and to align with other WebJobs extensions. Other extensions have namespaces and assembly names starting with `Microsoft` (the company name is required). I don't expect this to impact client applications as those application project files reference packages instead of assemblies, and the package name is not changed. Also, they will receive the update only on opting to update to next version of the extension package.

- Deleted two files not in use.
- Add VS Code file information in `.gitignore`.

WebJobs SDK code for reference: [1](https://github.com/Azure/azure-webjobs-sdk/blob/88ed7806c63d9c589ecf1a44ec7575f1e72384e5/src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj#L11-L12), [2](https://github.com/Azure/azure-webjobs-sdk/blob/88ed7806c63d9c589ecf1a44ec7575f1e72384e5/test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/WebJobs.Extensions.Storage.UnitTests.csproj#L6-L7).


src/Config/RabbitMQWebJobsBuilderExtensions.cs
```diff
@@ -3,12 +3,10 @@
 
 using System;
 using Microsoft.Azure.WebJobs;
-using Microsoft.Azure.WebJobs.Extensions;
-using Microsoft.Azure.WebJobs.Extensions.RabbitMQ;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 
-namespace Microsoft.Extensions.Hosting
+namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
```
JatinSanghvi: Should revert this change.

v-shenoy (changes requested):
I think the filenames starting with **Webjobs.Extensions.RabbitMQ** should also be changed to include **Microsoft.Azure**.

JatinSanghvi:
> I think the filenames starting with **Webjobs.Extensions.RabbitMQ** should also be changed to include **Microsoft.Azure**.

This is to keep the project file names smaller as we have in https://cs.github.com/Azure/azure-webjobs-sdk?q=path%3AWebjobs*csproj or in this repository. The root-namespace and assembly names are selected same to be project file names by default, but since both differ, we are adding `RootNamespace` and `AssemblyName` XML elements to the CSPROJ files. The projects in WebJobs SDK repo also does the same. I think that RabbitMQ extension followed the same convention as in WebJobs SDK repo but did not add the two XML elements as the original repo did.

v-shenoy (approved):

PR #190, opened 2022-03-17
Enable Automatic Recovery

aaguilartablada:
When connection errors occur, the RabbitMQ client does NOT recover automatically unless we set AutomaticRecoveryEnabled to true in our connection factory.

https://www.rabbitmq.com/dotnet-api-guide.html#connection-recovery


JatinSanghvi:
Thanks @aaguilartablada. I was not aware of the functionality and the `ConnectionFactory` option. That being said, [the default value of the option is set to true in the RabbitMQ client library](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/d4b49cf9ffc2959398c499884a04e0fc333dc2af/projects/RabbitMQ.Client/client/api/ConnectionFactory.cs#L155-L159) so this change would be a no-op. I will be closing the PR. Please reopen it and let me know if you think I am misjudging. I wanted to clarify that we really welcome and value contributions so please feel free if you have anything to add in future.

PR #193, opened 2022-03-22
Enable source link and deterministic build

JatinSanghvi:
Below screenshot shows the differences in NuGet package metadata as viewed on NuGet package explorer, between existing package v2.0.0-preview and the future one. Please ignore the version number and repository URL. They will be correctly set when a new tagged release will be created.

image

- Enabled Code Analysis.
- Replaced ruleset file with `.editorconfig`. This is modern way to specify the code quality and style rules.
- Retaining third party analyzer, StyleCop, as it has host of style checks, not available in default .NET analyzers.
- Updated the code to remove code analysis reported issues.
- Added icon. I see [three different images](https://www.nuget.org/packages?q=Microsoft.Azure.WebJobs.Extensions.) in use, so picked the one most appropriate in my opinion.

In next PR, the code will be further updated to remove some of the code analysis suppressions that are temporarily added in the .editorconfig file.


extension/.editorconfig
```diff
@@ -0,0 +1,147 @@
...
+dotnet_diagnostic.CA1014.severity = none # Mark assemblies with CLSCompliant
+dotnet_diagnostic.CA1308.severity = none # Normalize strings to uppercase
+dotnet_diagnostic.CA1716.severity = none # Identifiers should not match keywords
+dotnet_diagnostic.IDE0058.severity = none # Remove unnecessary expression value
+dotnet_diagnostic.IDE0130.severity = none # Namespace does not match folder structure
+dotnet_diagnostic.SA0001.severity = none # XML comment analysis is disabled (to be removed)
+dotnet_diagnostic.SA1101.severity = none # Prefix local calls with this
+dotnet_diagnostic.SA1202.severity = none # Elements must be ordered by access (to be removed)
+dotnet_diagnostic.SA1309.severity = none # Field name should not begin with an underscore (to be removed)
+
```
JatinSanghvi, line 82: Code analysis suppression were present in ruleset file earlier. I have marked some of the suppressions as `to be removed` in the comments as the code will be further updated in future PR to match the style guide.

extension/WebJobs.Extensions.RabbitMQ.Samples/WebJobs.Extensions.RabbitMQ.Samples.csproj
```diff
@@ -6,6 +6,8 @@
     <LangVersion>latest</LangVersion>
     <AssemblyName>Microsoft.Azure.WebJobs.Extensions.RabbitMQ.Samples</AssemblyName>
     <RootNamespace>Microsoft.Azure.WebJobs.Extensions.RabbitMQ.Samples</RootNamespace>
+    <IsPackable>false</IsPackable>
+    <EnableNETAnalyzers>false</EnableNETAnalyzers>
```
JatinSanghvi: The test and sample projects are targeting .NET 6. For .NET 6, if `.editorconfig` is present in project folder or in one of its parent folders, the .NET analyzers get kicked in automatically without any additional steps. For now, I need to disable the analysis, else otherwise the build will fail due to code-analysis issues. I will try to update code in both test and sample projects, so that .NET analyzer can be enabled, before publishing this draft PR.

extension/WebJobs.Extensions.RabbitMQ/Bindings/RabbitMQAsyncCollector.cs
```diff
@@ -17,16 +17,14 @@ internal class RabbitMQAsyncCollector : IAsyncCollector<ReadOnlyMemory<byte>>
...
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
-            _context = context ?? throw new ArgumentNullException(nameof(context));
-            if (_context.Service == null)
-            {
-                throw new ArgumentNullException("context.service");
-            }
+            _ = context ?? throw new ArgumentNullException(nameof(context));
+            _ = context.Service ?? throw new ArgumentException("Value cannot be null. Parameter name: context.Service");
+            _context = context;
```
JatinSanghvi, line 22: These are all changes to fix code quality and code style analysis reported issues.

extension/WebJobs.Extensions.RabbitMQ/Services/RabbitMQService.cs
```diff
@@ -9,21 +9,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
...
     internal sealed class RabbitMQService : IRabbitMQService
     {
-        private readonly IRabbitMQModel _rabbitMQModel;
-        private readonly IModel _model;
-        private readonly string _connectionString;
         private readonly string _queueName;
-        private readonly bool _disableCertificateValidation;
-        private readonly object _publishBatchLock;
-
-        private IBasicPublishBatch _batch;
```
JatinSanghvi: Code analyzer does not like public properties wrapping over private fields.

extension/WebJobs.Extensions.RabbitMQ/WebJobs.Extensions.RabbitMQ.csproj
```diff
@@ -1,58 +1,65 @@
...
-    <StartupObject />
     <SignAssembly>false</SignAssembly>
     <AssemblyOriginatorKeyFile>PublicKey.snk</AssemblyOriginatorKeyFile>
     <DelaySign>true</DelaySign>
+    <LangVersion>latest</LangVersion>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <AnalysisMode>All</AnalysisMode>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <!-- <Nullable>enable</Nullable> -->
```
JatinSanghvi: Will try to enable [Nullable option](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/language#nullable) before publishing the PR.

JatinSanghvi: Skipping this change. It would take effort to satisfy nullability checks. Also, there is chance of getting breaking changes introduced.

extension/WebJobs.Extensions.RabbitMQ.Samples/RabbitMQSamples.cs
```diff
@@ -120,14 +120,9 @@ public static class RabbitMQSamples
 
         public class TestClass
         {
-            private readonly int _x;
-            private readonly int _y;
+            public int X { get; set; }
```
JatinSanghvi, line 124: Private fields are ignored during serialisation and de-serialisation. Serializing TestClass object was yielding an empty JSON object `{}`.

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQAsyncCollectorTests.cs
```diff
@@ -16,7 +16,9 @@ public class RabbitMQAsyncCollectorTests
         [Fact]
         public async Task AddAsync_AddsMessagesToQueue()
         {
-            var batchLock = new object();
+#pragma warning disable SA1000
+            object batchLock = new();
+#pragma warning restore SA1000
```
JatinSanghvi, line 21: Disabled warning as StyleCop was reporting error for valid C# 9 syntax. I can check afterwards if there is a beta/future version of StyleCop in making that will understand the new syntax (pun intended).

extension/WebJobs.Extensions.RabbitMQ/Config/RabbitMQExtensionConfigProvider.cs
```diff
@@ -37,12 +37,11 @@ public RabbitMQExtensionConfigProvider(IOptions<RabbitMQOptions> options, INameR
...
-            if (context == null)
-            {
-                throw new ArgumentNullException("context");
-            }
+            _ = context ?? throw new ArgumentNullException(nameof(context));
 
-            var rule = context.AddBindingRule<RabbitMQAttribute>();
+#pragma warning disable 0618
+            FluentBindingRule<RabbitMQAttribute> rule = context.AddBindingRule<RabbitMQAttribute>();
+#pragma warning restore 0618
```
JatinSanghvi, line 44: `FluentBindingRule` seems to be obsolete. I will check sometime in future why this is obsoleted and if there's a non-obsoleted replacement for it, I will switch the variable type accordingly.

extension/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "editor.formatOnSave": true
+}
```
v-shenoy: I think we should have the linter add / flag missing new lines at the end of every file (not just for code files).

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQClientBuilderTests.cs
```diff
@@ -23,8 +23,8 @@ public void Opens_Connection()
             mockServiceFactory.Setup(m => m.CreateService(It.IsAny<string>(), false)).Returns(new Mock<IRabbitMQService>().Object);
             RabbitMQAttribute attr = GetTestAttribute();
 
-            RabbitMQClientBuilder clientBuilder = new RabbitMQClientBuilder(config, options);
-            var model = clientBuilder.Convert(attr);
+            var clientBuilder = new RabbitMQClientBuilder(config, options);
+            IModel model = clientBuilder.Convert(attr);
```
v-shenoy, line 27: When are we agreeing to use / not use `var` as per the style guidelines?

extension/.vscode/settings.json
JatinSanghvi: Keeping it the way it is, as this is how Visual Studio Code formats the JSON files. If we add a new line at the end of JSON file and press Alt+Shift+F, the ending new line character gets removed.

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQClientBuilderTests.cs
JatinSanghvi: Use `var` when the type is evident e.g., when we are new-ing a class object. Use actual type otherwise e.g., for storing function return values. The styles are enforced to make choice of var vs no-var consistent across all files.

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQListenerTests.cs
```diff
@@ -43,12 +43,12 @@ public RabbitMQListenerTests()
         [Fact]
         public void CreatesHeadersAndRepublishes()
         {
-            var queueName = "blah";
+            string queueName = "blah";
```
v-shenoy, line 46: In reference to the previous question, is there a reason to remove `var` here?

extension/.vscode/settings.json
v-shenoy:
```json 
{
   "files.insertFinalNewline": true,
    "files.trimFinalNewlines": true,
}
```

VS Code also has these settings that ensures that there is exactly one new line at the end of the file.

JatinSanghvi: Thanks for suggesting. I have incorporated the change and updated all files not having final new line.

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQListenerTests.cs
JatinSanghvi: The `var`s were not uniformly applied in the source code when literal values were being assigned to variable. We had to choose between var or explicit types. The style rules are set such that only when the type-name is repeated on both left and right side of the assignment operator, we will use var otherwise won't.

v-shenoy: Thanks for the clarification.

v-shenoy (approved):
LGTM!

PR #194, opened 2022-04-19
Remove StyleCop suppressions and resolve style issues

JatinSanghvi:
Adhering to default StyleCop recommendations i.e., using `this.` in front of the instance variables. This way is more consistent since otherwise we use variables with names like `_varName` for private instance fields, and names like `VarName` for public and private static fields and public instance properties. When referencing one of these variables in rest of the class, having `this.` prefixed makes it clear to the reader whether instance or static or local variable is being accessed. In that case, the underscore prefix becomes irrelevant. Similar justification can be found in StyleCop documentation [here](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1309.md#rule-description) and [here](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1101.md#rule-description).

### DIFF
--- a/extension/.editorconfig
+++ b/extension/.editorconfig
@@ -27,10 +27,10 @@ charset = utf-8-bom
 # Organize usings
 dotnet_sort_system_directives_first = true
 # this. preferences
-dotnet_style_qualification_for_field = false:silent
-dotnet_style_qualification_for_property = false:silent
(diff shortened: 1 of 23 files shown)


v-shenoy (approved):
LGTM!

PR #198, opened 2022-06-24
Fix exceptions when connection string is not specified

JatinSanghvi:
If the connection-string setting name is not explicitly specified by user in RabbitMQ trigger binding, the method `ResolveConnectionString` internally throws `ArgumentNullException`. To fix the StyleCop errors, I had switched catching of generic exception `Exception` to `InvalidOperationException` which caused this regression. I am unable to remember why this specific exception was chosen but it should be by going through the code of `GetConnectionStringOrSetting`.

Instead of going back to generic `Exception`, I found it more apt to directly check if the connection string key is not specified explicitly by the user. WebJobs SDK code also calls `GetConnectionStringOrSetting` at few places but does not handle any exception. Note that the problem is only with trigger bindings; it does not affect the output binding functionality.


v-shenoy (approved):
LGTM.

PR #201, opened 2022-07-07
Added distributed tracing instrumentation

t-manyadav:
Added distributed tracing instrumentation. Users can now collect the distributed trace telemetry to diagnose application issues when needed. 

Message producers need to pass "traceparent" id in headers of message while publishing the message to have distributed tracing, otherwise created activity will be treated as root activity.
Please note: TraceParentId needs to be in [W3C trace context](https://www.w3.org/TR/trace-context/) format.

Since distributed tracing is not fully supported yet, users need to add a file (say starup.cs) in root folder of function app and do openTelemetry configuration.

Follow these resources for more details:
 [Adding distributed tracing instrumentation](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-instrumentation-walkthroughs)
[openTelemetry Configuration .NET](https://opentelemetry.io/docs/instrumentation/net/getting-started/)

Sample code for startup.cs file:
image

### DIFF
--- a/extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
+++ b/extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
     internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTriggerMetrics>
     {
 #pragma warning disable SA1000
-        private static readonly ActivitySource Source = new("Microsoft.Azure.WebJobs.Extensions.RabbitMQ");
+        private static readonly ActivitySource ActivitySource = new("Microsoft.Azure.WebJobs.Extensions.RabbitMQ");
 #pragma warning restore SA1000
         private readonly ITriggeredFunctionExecutor executor;
         private readonly string queueName;
@@ -86,7 +86,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 
             this.consumer.Received += async (model, ea) =>
             {
-                Activity activity = StartActivity(ea); // create activity object
+                using Activity activity = StartActivity(ea);
 
                 var input = new TriggeredFunctionData() { TriggerValue = ea };
                 FunctionResult result = await this.executor.TryExecuteAsync(input, cancellationToken).ConfigureAwait(false);
@@ -106,8 +106,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
                         this.RepublishMessages(ea);
                     }
                 }
-
-                activity?.Stop();
             };
 
             this.consumerTag = this.rabbitMQModel.BasicConsume(queue: this.queueName, autoAck: false, consumer: this.consumer);
@@ -165,22 +163,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
             Activity activity;
             if (ea.BasicProperties.Headers != null && ea.BasicProperties.Headers.ContainsKey("traceparent"))
             {
-                // check if traceId present in header
                 byte[] traceParentIdInBytes = ea.BasicProperties.Headers["traceparent"] as byte[];
                 string traceparentId = Encoding.Default.GetString(traceParentIdInBytes);
-                activity = Source.StartActivity("RabbitMQ.function.trigger", ActivityKind.Consumer, traceparentId);
+                activity = ActivitySource.StartActivity("Trigger", ActivityKind.Consumer, traceparentId);
             }
             else
             {
-                // create new traceId, StartActivity will create and start new activity
-                activity = Source.StartActivity("RabbitMQ.function.trigger", ActivityKind.Server);
+                activity = ActivitySource.StartActivity("Trigger", ActivityKind.Server);
                 if (ea.BasicProperties.Headers == null)
                 {
                     ea.BasicProperties.Headers = new Dictionary<string, object>();
                 }
 
-                byte[] traceParentIdInBytes = Encoding.Default.GetBytes(activity?.Id);
-                ea.BasicProperties.Headers["traceparent"] = traceParentIdInBytes; // add trace-id to header
+                byte[] traceParentIdInBytes = Encoding.Default.GetBytes(activity.Id);
+                ea.BasicProperties.Headers["traceparent"] = traceParentIdInBytes;
             }
 
             return activity;
--- a/extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
+++ b/extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Executors;
@@ -20,6 +21,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 {
     internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTriggerMetrics>
     {
+#pragma warning disable SA1000
+        private static readonly ActivitySource Source = new("Microsoft.Azure.WebJobs.Extensions.RabbitMQ");
+#pragma warning restore SA1000
         private readonly ITriggeredFunctionExecutor executor;
         private readonly string queueName;
         private readonly ushort prefetchCount;
@@ -82,6 +86,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 
             this.consumer.Received += async (model, ea) =>
             {
+                Activity activity = StartActivity(ea); // create activity object
+
                 var input = new TriggeredFunctionData() { TriggerValue = ea };
                 FunctionResult result = await this.executor.TryExecuteAsync(input, cancellationToken).ConfigureAwait(false);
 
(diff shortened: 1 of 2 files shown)


extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -82,6 +86,8 @@ public Task StartAsync(CancellationToken cancellationToken)
 
             this.consumer.Received += async (model, ea) =>
             {
+                Activity activity = StartActivity(ea); // create activity object
+
```
JatinSanghvi, line 90: The comments are not adding to the information. Can you remove all of them?

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -82,6 +86,8 @@ public Task StartAsync(CancellationToken cancellationToken)
...
                 var input = new TriggeredFunctionData() { TriggerValue = ea };
```
JatinSanghvi, line 91: Can you use the using statement on Activity object? The `Dispose` method will automatically [call `Stop` on the Activity](https://github.com/dotnet/runtime/blob/55e2378d86841ec766ee21d5e504d7724c39b53b/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L1009-L1012). The using statement can handle `null` Activity objects.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -20,6 +21,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 {
     internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTriggerMetrics>
     {
+#pragma warning disable SA1000
+        private static readonly ActivitySource Source = new("Microsoft.Azure.WebJobs.Extensions.RabbitMQ");
```
JatinSanghvi: `Source` is too generic to be used as class member variable name. Rename it to `ActivitySource`.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -152,6 +160,32 @@ public ScaleStatus GetScaleStatus(ScaleStatusContext<RabbitMQTriggerMetrics> con
...
 
+        internal static Activity StartActivity(BasicDeliverEventArgs ea)
+        {
+            Activity activity;
+            if (ea.BasicProperties.Headers != null && ea.BasicProperties.Headers.ContainsKey("traceparent"))
+            {
+                // check if traceId present in header
+                byte[] traceParentIdInBytes = ea.BasicProperties.Headers["traceparent"] as byte[];
+                string traceparentId = Encoding.Default.GetString(traceParentIdInBytes);
+                activity = Source.StartActivity("RabbitMQ.function.trigger", ActivityKind.Consumer, traceparentId);
```
JatinSanghvi: Should we have the Activity name as `Trigger` only since the Activity Source gives rest of the information?

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -152,6 +160,32 @@ public ScaleStatus GetScaleStatus(ScaleStatusContext<RabbitMQTriggerMetrics> con
...
+            else
+            {
+                // create new traceId, StartActivity will create and start new activity
+                activity = Source.StartActivity("RabbitMQ.function.trigger", ActivityKind.Server);
+                if (ea.BasicProperties.Headers == null)
+                {
+                    ea.BasicProperties.Headers = new Dictionary<string, object>();
+                }
+
+                byte[] traceParentIdInBytes = Encoding.Default.GetBytes(activity?.Id);
```
JatinSanghvi: Change to `activity.Id` as it gives impression that `activity` can possibly be `null`.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -20,6 +21,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 {
     internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTriggerMetrics>
     {
+#pragma warning disable SA1000
```
v-shenoy, line 24: Use class name with new instead of disabling the style check.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -20,6 +21,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
...
+        private static readonly ActivitySource ActivitySource = new("Microsoft.Azure.WebJobs.Extensions.RabbitMQ");
```
v-shenoy, line 25: Use class name with new instead of disabling the style check.

JatinSanghvi: Doing that makes C# analyzer recommend the latest syntax, so there is no syntax that can silence both analyzers. The issue is fixed in StyleCop analyzer. I will check again why we cannot have the latest StyleCop package. The last thing we can try is to disable the warning in .editorconfig but that will disable all warnings where whitespaces are not added after keywords.

v-shenoy: That's weird with both analyzers recommending conflicting styles.

Yeah, I had rather not disable `SA1000` globally just to get over this one specific case. Are you sure this has been fixed in the StyleCop analyzer? 

[SA1000.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1000.md) doesn't mention it.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -152,6 +158,30 @@ public ScaleStatus GetScaleStatus(ScaleStatusContext<RabbitMQTriggerMetrics> con
             return this.GetScaleStatusCore(context.WorkerCount, context.Metrics?.ToArray());
         }
 
+        internal static Activity StartActivity(BasicDeliverEventArgs ea)
+        {
+            Activity activity;
+            if (ea.BasicProperties.Headers != null && ea.BasicProperties.Headers.ContainsKey("traceparent"))
+            {
+                byte[] traceParentIdInBytes = ea.BasicProperties.Headers["traceparent"] as byte[];
+                string traceparentId = Encoding.Default.GetString(traceParentIdInBytes);
```
JatinSanghvi: Why do we need to encode/decode the header value? I see RabbitMQ client encodes the string in UTF-8 by itself.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -152,6 +158,30 @@ public ScaleStatus GetScaleStatus(ScaleStatusContext<RabbitMQTriggerMetrics> con
...
+                activity = ActivitySource.StartActivity("Trigger", ActivityKind.Consumer, traceparentId);
+            }
+            else
+            {
+                activity = ActivitySource.StartActivity("Trigger", ActivityKind.Server);
+                if (ea.BasicProperties.Headers == null)
```
JatinSanghvi: Can you check if the [null-coalescing assignment operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator) from C# 8.0 works here?

```suggestion
ea.BasicProperties.Headers ??= new Dictionary<string, object>();
```

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
JatinSanghvi: It's fixed in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3187

v-shenoy: Yeah, this has been fixed in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/commit/3874af7bcc5b3b79c628f824ea88c185bc5bc374

Guess the doc file just hasn't been updated.

Edit : Whoops, commented at the same time.

JatinSanghvi: So StyleCop analyzer only has beta packages released since last three years. Taking dependency on the latest pre-release package should fix this problem.

v-shenoy: As per the PR thread, they recommend using the `1.2.xxx-beta` version of their NuGet package. Kind of annoying given the 1.2 has been beta for well over years now.

v-shenoy: Guess so. They really should release the official version `1.2`.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
t-manyadav: When header value is string type, RabbitMQ convert it into byte array.

JatinSanghvi (approved):

JatinSanghvi:
Thanks @t-manyadav for your contribution 👍

PR #215, opened 2022-09-30
Fix message body corruption on republish

### LINKED ISSUE #211: ERROR: corrupted messages continues on 2.0.1
Hello.

The problem described in https://github.com/Azure/azure-functions-rabbitmq-extension/issues/207 still occurs. I have tested the same dummy function with the version 2.0.1 and I still see corrupted messages.

I have tested it over Function Host 4.1.1 and 4.11.0 with no difference.

JatinSanghvi:
- Closes #211

### DIFF
--- a/extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
+++ b/extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
@@ -21,7 +21,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ
 {
     internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTriggerMetrics>
     {
+        private const string RequeueCountHeaderName = "x-ms-rabbitmq-requeuecount";
+
         private static readonly ActivitySource ActivitySource = new("Microsoft.Azure.WebJobs.Extensions.RabbitMQ");
+
         private readonly ITriggeredFunctionExecutor executor;
         private readonly string queueName;
         private readonly ushort prefetchCount;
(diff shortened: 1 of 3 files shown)


extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -89,27 +92,35 @@ public Task StartAsync(CancellationToken cancellationToken)
...
-                    if (ea.BasicProperties.Headers == null || !ea.BasicProperties.Headers.ContainsKey(Constants.RequeueCount))
-                    {
-                        this.CreateHeadersAndRepublish(ea);
-                    }
-                    else
+                    ea.BasicProperties.Headers ??= new Dictionary<string, object>();
+                    ea.BasicProperties.Headers.TryGetValue(RequeueCountHeaderName, out object headerValue);
+                    int requeueCount = Convert.ToInt32(headerValue, CultureInfo.InvariantCulture) + 1;
+
+                    if (requeueCount >= 5)
```
JatinSanghvi, line 101: Behavior is changed here. Same message will trigger the user function 5 times instead of 6 times. There was code duplication between `CreateHeadersAndRepublish` and `RepublishMessages` requiring the same fix to be added to both methods, hence trying to simplify the logic and removing redundancy.

v-shenoy, line 101: Looks good overall, but had a question. Is the choice of 5 as the requeue count limit arbitrary here?

JatinSanghvi: Let me check if there is documentation on this, and if it is consistent or differs with each extension.

JatinSanghvi: There is some documentation on retries on <https://learn.microsoft.com/azure/azure-functions/functions-bindings-error-pages#retries>, but the retry counts are not documented. The Kafka extension Readme file mentions the default to be 2, which Cosmos DB does not seem to have any retry in place. For many extensions, the retry count is configurable through 'max retry count' property in host.json or via `FixedDelayRetry` or `ExponentialBackoffRetry` attribute on triggered function.

Note that retry via republish is a better option that the function executor itself invoking the function many times, as in the former case, the republished message is likely to end up on a different function app instance if multiple of them are running at a time.

v-shenoy (approved):

PR #218, opened 2022-10-14
Fix message body corruption on republish

### LINKED ISSUE #211: ERROR: corrupted messages continues on 2.0.1

JatinSanghvi:
The fix is based on the suggestion made in #211.

- Fixes #211.

### DIFF
--- a/extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
+++ b/extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
@@ -85,38 +85,44 @@ internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTrigge
         this.rabbitMQModel.BasicQos(0, this.prefetchCount, false);  // Non zero prefetchSize doesn't work (tested upto 5.2.0) and will throw NOT_IMPLEMENTED exception
         this.consumer = new EventingBasicConsumer(this.rabbitMQModel.Model);
 
-        this.consumer.Received += async (model, ea) =>
+        this.consumer.Received += async (model, args) =>
         {
-            using Activity activity = StartActivity(ea);
+            // The RabbitMQ client rents an array from the ArrayPool to hold a copy of the message body, and passes it
+            // to the listener. Once all event handlers are executed, the array is returned back to the pool so that the
+            // memory can be reused for future messages for that connection. However, since our event handler is async,
+            // the very first await statement i.e. the call to TryExecuteAsync below causes the event handler invocation
+            // to complete and lets the RabbitMQ client release the memory. This led to message body corruption when the
+            // message is republished (see: https://github.com/Azure/azure-functions-rabbitmq-extension/issues/211).
+            //
+            // We chose to copy the message body instead of having a new 'args' object as there is only one event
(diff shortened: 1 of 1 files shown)


extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -85,38 +85,45 @@ public Task StartAsync(CancellationToken cancellationToken)
...
+            // to the listener. Once all event handlers are executed, the array is returned back to the pool so that the
+            // memory can be reused for future messages for that connection. However, since our event handler is async,
+            // the very first await statement i.e. the call to TryExecuteAsync causes the event handler invocation to
+            // complete and lets the RabbitMQ client reclaim the memory. This led to message body corruption when the
+            // message is republished (see: https://github.com/Azure/azure-functions-rabbitmq-extension/issues/211).
+            //
+            // Since the same 'args' argument is passed to all event handlers, replacing it with a local copy (with
+            // message body copied) will ensure that the other event handlers (in case they are present) will receive
+            // exactly the same args as it was composed by the RabbitMQ client.
+            args = new BasicDeliverEventArgs(args.ConsumerTag, args.DeliveryTag, args.Redelivered, args.Exchange, args.RoutingKey, args.BasicProperties, args.Body.ToArray());
```
yvanoers: `args.BasicProperties` is not deep-copied. If `args.BasicProperties.Headers` is already initialized when passed to the handler, line 122 will modify the original Headers object, not a copy.

JatinSanghvi: I know. There are a lot of properties inside class `BasicProperties` and I doubt we will benefit by having a piece of code that performs the deep copy. The code always had this issue since the beginning, so in that sense, deep copying would fall out of scope of the fix we are trying to make. I am now thinking that we are probably overdoing by making a copy of args, as the RabbitMQ extension would not be sharing the consumer object with rest of the extension code, so there's no chance of multiple handlers getting invoked in its current state.

yvanoers: I just pointed it out because the code was changed to copy the entirety of `args`.
Because this is indeed besides the point of the fix, I submitted it as a comment, not a request for change.

JatinSanghvi: Sure, thanks. I will add a comment explaining the reason why a deep copy is not required currently. As a side note, the ideal way would be to make two copies of args in the beginning, one to be passed to the user function, so any change the function makes in the BasicDeliverEventArgs trigger parameter (if that is the type the trigger is bound to) does not affect the republished message, and another one to be used in rest of the extension code including the republishing logic.

v-shenoy (approved):
I went through the linked issue trying to understand what was causing the corruption. Good work on resolving with help from the community!

LGTM.

PR #223, opened 2022-11-03
Refactor class `RabbitMQListener`

JatinSanghvi:


extension/WebJobs.Extensions.RabbitMQ.Tests/Trigger/RabbitMQListenerTests.cs
```diff
@@ -0,0 +1,263 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
JatinSanghvi, line 1: Moved file `extension/WebJobs.Extensions.RabbitMQ.Tests\RabbitMQListnerTests.cs` to `extension/WebJobs.Extensions.RabbitMQ.Tests\Trigger\RabbitMQListnerTests.cs` but also replaces all tests.

The previously commented out tests that used to test for retry logic are not added, since the implementation for mocking classes turned to be complicating both test file and the Listener class.

extension/WebJobs.Extensions.RabbitMQ/Services/IRabbitMQService.cs
```diff
@@ -7,8 +7,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ;
 
 public interface IRabbitMQService
 {
-    IRabbitMQModel RabbitMQModel { get; }
```
JatinSanghvi, line 10: `IRabbitMQModel` was redundant wrapper around `IModel`. I am not sure what it accomplished since `IModel` is itself an interface and fully mockable for tests. Also `IRabbitMQService` was keeping reference to both `IModel` and wrapper of same class `IRabbitMQModel` and on different times, it called one of the two properties, which totally looked illogical.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQActivitySource.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
JatinSanghvi, line 1: May be an UTF-8 BOM character. I am not reverting this change.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -20,126 +19,134 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ;
...
+    private readonly ILogger logger;
     private readonly string queueName;
     private readonly ushort prefetchCount;
-    private readonly IRabbitMQService service;
-    private readonly ILogger logger;
-    private readonly string functionId;
-    private readonly IRabbitMQModel rabbitMQModel;
+    private readonly string logDetails;
 
-    private EventingBasicConsumer consumer;
```
JatinSanghvi, line 33: Making this a local variable.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -20,126 +19,134 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ;
...
+    private int listenerState = ListenerNotStarted;
     private string consumerTag;
-    private bool disposed;
```
JatinSanghvi, line 35: Replaced `disposed` and `started` with stricter thread-safe checks. This pattern can also be found in few other extensions.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -20,126 +19,134 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ;
...
-            // The RabbitMQ client rents an array from the ArrayPool to hold a copy of the message body, and passes it
-            // to the listener. Once all event handlers are executed, the array is returned back to the pool so that the
-            // memory can be reused for future messages for that connection. However, since our event handler is async,
-            // the very first await statement i.e. the call to TryExecuteAsync below causes the event handler invocation
-            // to complete and lets the RabbitMQ client release the memory. This led to message body corruption when the
-            // message is republished (see: https://github.com/Azure/azure-functions-rabbitmq-extension/issues/211).
-            //
-            // We chose to copy the message body instead of having a new 'args' object as there is only one event
-            // handler registered for the consumer so there should be no side-effects.
-            args.Body = args.Body.ToArray();
```
JatinSanghvi, line 96: I found that copying message body is not required if we replace `EventingBasicConsumer` with `AsyncEventingBasicConsumer`. Please check my test code: https://gist.github.com/JatinSanghvi/0993825524957bb3f36f207ac7d4a676. The code copies the implementation of `AsyncEventingBasicConsumer` from RabbitMQ client.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -170,92 +178,98 @@ public ScaleStatus GetScaleStatus(ScaleStatusContext<RabbitMQTriggerMetrics> con
...
+    /// <summary>
+    /// Returns scale recommendation i.e. whether to scale in or out the host application. The recommendation is
+    /// made based on both the latest metrics and the trend of increase or decrease in the the number of messages in
+    /// Ready state in the queue. In all of the calculations, it is attempted to keep the number of workers minimum
+    /// while also ensuring that the message count per worker stays under the maximum limit.
+    /// </summary>
+    /// <param name="workerCount">The current worker count for the host application.</param>
+    /// <param name="metrics">The collection of metrics samples to make the scale decision.</param>
+    private ScaleStatus GetScaleStatusCore(int workerCount, RabbitMQTriggerMetrics[] metrics)
     {
```
JatinSanghvi, line 190: I have replaced the primitive scaling logic with one that I wrote for SQL extension. There were couple of issues with the existing implementation, primarily it used to scale out too aggressively.

v-shenoy (dismissed):

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -20,126 +19,134 @@ namespace Microsoft.Azure.WebJobs.Extensions.RabbitMQ;
 
 internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTriggerMetrics>
 {
+    private const int ListenerNotStarted = 0;
+    private const int ListenerStarting = 1;
+    private const int ListenerStarted = 2;
+    private const int ListenerStopping = 3;
+    private const int ListenerStopped = 4;
```
v-shenoy, line 26: Maybe replace with enum?

v-shenoy (commented):
Accidentally missed a file due to GitHub UI issue, will take a look again later by EOD.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
JatinSanghvi: It's not possible to use `Interlocked.CompareExchange` on enum values without workarounds. There is a GitHub issue on this: https://github.com/dotnet/runtime/issues/65184.

v-shenoy: I somewhat figured when I saw that later in the file. Looks like the cleanest way to do it without some inane workaround around `Interlocked.CompareExchange`, which is not really worth it.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
v-shenoy: This is not complete implementation of runtime-driven scaling, right? Just an improvement on the current implementation, and we will need another PR for runtime-driven scaling?

v-shenoy (dismissed):

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
JatinSanghvi: This is actually runtime driven scaling logic. We do not have RabbitMQ scaling logic encoded  inside Scale controller which prevents scaling support for consumption tier, and on dedicated tiers unless customer explicitly enable runtime driven scaling. By the way, I still need to test the update scaling implementation on real service. I will complete the PR after ensuring that scaling works as expected.

v-shenoy: It's really great if the scaling works without hiccups. We should get the extensions bundles updated after that.

JatinSanghvi: There was a bug but otherwise I could ensure that scaling works fine i.e., the number of instances increases on high load, it stays consistent with expected worker-count for continuous load and goes back to one (Elastic Premium behavior) when the queue becomes idle. Please review the new commit.

v-shenoy (approved):

PR #229, opened 2023-07-17
Update CODEOWNERS

v-shenoy:


JatinSanghvi (commented):
Can you add another owner, maybe Lakshmi, so that the review is not stuck for one person. For example, if we need to change code owners again in the distant future.

JatinSanghvi (approved):

PR #233, opened 2024-04-26
Weekly Sunday build for the CI

vivekjilla:


azure-pipelines.yml
```diff
@@ -25,6 +33,7 @@ variables:
   prefix: $[format('0.{0:yyyy}.{0:MMdd}', pipeline.startTime)]
   version: $[format('{0}.{1}', variables.prefix, counter(variables.prefix, 1))] # e.g. 0.2001.0203.4
   fileVersion: $[variables.version]
+  Codeql.Enabled: true
```
jainharsh98, line 36: Will codeql artifacts be uploaded for every pr build as well? If yes, Should we prevent that?

vivekjilla: it won't. looks like it gets uploaded only for default branch

jainharsh98 (approved):
LGTM

v-shenoy (approved):

PR #254, opened 2025-04-25
[Final] Migrate to 1ESPT pipelines

aloiva:


vivekjilla (dismissed):

eng/cd/official-release.yml
```diff
@@ -1,6 +1,19 @@
...
-  - template: /eng/cd/templates/release-variables.yml@self
+
+parameters:
+  - name: nugetPackageName
+    type: string
+    displayName: "Base name of the Nuget package"
+    default: "Microsoft.Azure.WebJobs.Extensions.RabbitMQ"
+  - name: mavenPackageName
+    type: string
+    displayName: "Base name of the Maven package"
```
manikantanallagatla, line 10: nit: we can add (ex.) in these displaynames as well like packageversion

manikantanallagatla (approved):

eng/cd/official-release.yml
aloiva: ideally, they shouldn't be edited since there is only one name possible so i didn't put in any example format. they'd be left at default.

Issue #151, opened 2021-02-16
Question: How do I manually Ack a message in dotnet core?


kentonbmax:
User Case: an EF error/timeout occurs while executing my function. Any other error occurs, I do not want to loose the message. Therefore I want to only manually Ack messages that have been successfully processed.

yojagad:
Needs some design exploration. Probably needs an approach similar to service bus where customers can specify messageHandlingOptions like below

```
{
  "version": "2.0",
  "extensions": {
    "rabbitMq": {
      "messageHandlerOptions": {
        "autoComplete": false
      }
    }
  }
}
```

DavidRobinsonDk:
This seems like a "must have" feature.
Queue handling without the possibility of abandoned messages with re-sends, doesn't really seem like queue handling. More sort of like "spray and pray" IMHO.

steveb85:
did this ever get worked on?

PR #255, opened 2025-05-09
support manualAck attribute for in-proc

ishank12:
As discussed in the issue https://github.com/Azure/azure-functions-rabbitmq-extension/issues/151

It is common scenario that message ack / reject would need to be done by the function code and not always by the host.
This PR is fixing that problem by introducing a new trigger attribute DisableAck. 
If this attribute is set to true, host would not perform acknowledgements for the processed message and function code need to do it.
The default value of this attribute is false, so existing apps would be working as is. 

Since the Ack/Reject needs to be done on the exact same channel instance which was created by the host, the host share this channel instance encapsulated in a class RabbitMQMessageActions and pass it as trigger binding.

below code is for reference how it can be used.
```
[FunctionName("Function1")]
public void Run([RabbitMQTrigger("test11", ConnectionStringSetting = "rabbitmqconnsetting", DisableAck = true)]string myQueueItem,
    string consumerTag,
    ulong deliveryTag,
    RabbitMQMessageActions messageActions,
    ILogger log)
{
    log.LogInformation($"C# Queue trigger function processed: {myQueueItem}, deliveryTag: {deliveryTag}");
    
    Console.WriteLine("Acknowleding...");
    messageActions.Acknowledge();
}
```

_Note:
Now, since the host is passing the channel instance to the function code, this is currently only supporting in-proc and C# functions only.
for non C# functions DisableAck should be always false._

### DIFF
--- a/extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
+++ b/extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
@@ -115,12 +115,12 @@ public class RabbitMQTriggerBindingTests
         await listener.StartAsync(CancellationToken.None);
 
         // Find the Consumer instance passed to RabbitMQService.Consume method
-        IInvocation basicConsumeInvocation = mockservice.Invocations
+        IInvocation consumeInvocation = mockservice.Invocations
             .FirstOrDefault(invocation => invocation.Method.Name == "Consume");
 
-        Assert.NotNull(basicConsumeInvocation);
+        Assert.NotNull(consumeInvocation);
 
-        var consumer = basicConsumeInvocation.Arguments[2] as AsyncEventingBasicConsumer;
+        var consumer = consumeInvocation.Arguments[2] as AsyncEventingBasicConsumer;
         Assert.NotNull(consumer);
 
         // Simulate message delivery
--- a/extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
+++ b/extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Moq;
 using RabbitMQ.Client;
 using RabbitMQ.Client.Events;
 using Xunit;
@@ -23,6 +24,7 @@ public class RabbitMQTriggerBindingTests
             ["RoutingKey"] = typeof(string),
             ["BasicProperties"] = typeof(IBasicProperties),
             ["Body"] = typeof(ReadOnlyMemory<byte>),
+            ["RabbitMQMessageActions"] = typeof(RabbitMQMessageActions),
         };
 
         IReadOnlyDictionary<string, Type> actualContract = RabbitMQTriggerBinding.CreateBindingDataContract();
@@ -44,6 +46,7 @@ public class RabbitMQTriggerBindingTests
 
         ReadOnlyMemory<byte> body = buffer;
         var eventArgs = new BasicDeliverEventArgs("ConsumerName", deliveryTag, false, "n/a", "QueueName", null, body);
+        var messageActions = new RabbitMQMessageActions(Mock.Of<IModel>());
 
         var data = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase)
         {
@@ -54,13 +57,42 @@ public class RabbitMQTriggerBindingTests
             ["Body"] = body,
             ["Exchange"] = eventArgs.Exchange,
             ["BasicProperties"] = eventArgs.BasicProperties,
+            ["RabbitMQMessageActions"] = messageActions,
         };
 
-        IReadOnlyDictionary<string, object> actualContract = RabbitMQTriggerBinding.CreateBindingData(eventArgs);
+        IReadOnlyDictionary<string, object> actualContract = RabbitMQTriggerBinding.CreateBindingData(eventArgs, messageActions);
 
         foreach (KeyValuePair<string, object> item in actualContract)
         {
             Assert.Equal(data[item.Key], item.Value);
         }
+    }
+
+    [Fact]
+    public void RabbitMQTriggerAttribute_ManualAck_DefaultsToFalse()
+    {
+        var attribute = new RabbitMQTriggerAttribute("test-queue");
+        bool manualAck = attribute.ManualAck;
+        Assert.False(manualAck, "ManualAck should default to false.");
+    }
+
+    [Fact]
+    public void RabbitMQTriggerBinding_CreateBindingData_IncludesRabbitMQMessageActions()
+    {
+        ulong deliveryTag = 1;
+
+        var rand = new Random();
+        byte[] buffer = new byte[10];
+        rand.NextBytes(buffer);
+
+        ReadOnlyMemory<byte> body = buffer;
+        var eventArgs = new BasicDeliverEventArgs("ConsumerName", deliveryTag, false, "n/a", "QueueName", null, body);
+        var messageActions = new RabbitMQMessageActions(Mock.Of<IModel>());
+
+        IReadOnlyDictionary<string, object> bindingData = RabbitMQTriggerBinding.CreateBindingData(eventArgs, messageActions);
+
+        // Assert
+        Assert.True(bindingData.ContainsKey("RabbitMQMessageActions"), "Binding data should include RabbitMQMessageActions.");
+        Assert.Equal(messageActions, bindingData["RabbitMQMessageActions"]);
     }
 }
--- a/extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
+++ b/extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
@@ -3,6 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Extensions.Logging;
 using Moq;
 using RabbitMQ.Client;
 using RabbitMQ.Client.Events;
@@ -94,5 +100,73 @@ public class RabbitMQTriggerBindingTests
         // Assert
         Assert.True(bindingData.ContainsKey("MessageActions"), "Binding data should include MessageActions.");
         Assert.Equal(messageActions, bindingData["MessageActions"]);
+    }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public async Task RabbitMQTrigger_ManualAck_BasicAckBehavior(bool manualAck)
+    {
+        // Arrange
+        var mockChannel = new Mock<IModel>();
+        var mockExecutor = new Mock<ITriggeredFunctionExecutor>();
+        var mockLogger = new Mock<ILogger>();
+        var mockDrainModeManager = new Mock<IDrainModeManager>();
+        var mockBasicProperties = new Mock<IBasicProperties>();
+
+        // Simulate successful function execution
+        mockExecutor
+            .Setup(executor => executor.TryExecuteAsync(It.IsAny<TriggeredFunctionData>(), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(new FunctionResult(true));
+
+        var listener = new RabbitMQListener(
+            mockChannel.Object,
+            mockExecutor.Object,
+            mockLogger.Object,
+            functionId: "test-function",
+            queueName: "test-queue",
+            manualAck: manualAck,
+            prefetchCount: 10,
+            drainModeManager: mockDrainModeManager.Object);
+
+        var eventArgs = new BasicDeliverEventArgs
+        {
+            DeliveryTag = 1,
+            Body = new ReadOnlyMemory<byte>([0x01, 0x02, 0x03]),
+            BasicProperties = mockBasicProperties.Object,
+        };
+
+        // Act
+        await listener.StartAsync(CancellationToken.None);
+
+        // Find the IBasicConsumer passed to BasicConsume
+        IInvocation basicConsumeInvocation = mockChannel.Invocations
+            .FirstOrDefault(invocation => invocation.Method.Name == "BasicConsume");
+
+        Assert.NotNull(basicConsumeInvocation);
+
+        // The third argument is the consumer
+        var consumer = basicConsumeInvocation.Arguments[6] as AsyncEventingBasicConsumer;
+        Assert.NotNull(consumer);
+
+        // Simulate message delivery
+        await consumer.HandleBasicDeliver(
+            consumerTag: "ctag",
+            deliveryTag: eventArgs.DeliveryTag,
+            redelivered: false,
+            exchange: string.Empty,
+            routingKey: string.Empty,
+            properties: eventArgs.BasicProperties,
+            body: eventArgs.Body.ToArray());
+
+        // Assert
+        if (manualAck)
+        {
+            mockChannel.Verify(channel => channel.BasicAck(It.IsAny<ulong>(), It.IsAny<bool>()), Times.Never, "BasicAck should not be called when ManualAck is true.");
+        }
+        else
+        {
+            mockChannel.Verify(channel => channel.BasicAck(It.IsAny<ulong>(), It.IsAny<bool>()), Times.AtLeastOnce, "BasicAck should be called when ManualAck is false.");
+        }
     }
 }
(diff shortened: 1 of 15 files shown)


vivekjilla:
Can you add detailed description of changes, testing done etc in PR description above, for future reference

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -133,10 +142,15 @@ async Task ReceivedHandler(object model, BasicDeliverEventArgs args)
...
-            }
 
-            // Acknowledge the existing message only after the message (in case of failure) is re-published.
-            this.channel.BasicAck(args.DeliveryTag, multiple: false);
+                // Acknowledge the existing message after the message is re-published.
+                this.channel.BasicAck(args.DeliveryTag, multiple: false);
+            }
+            else if (!this.manualAck)
+            {
+                // Acknowledge the existing message if manualAck is not set and function execution was successful.
```
vivekjilla, line 149: Shall we add a log when manualAck is true, calling out that `manualAck is true, so not acknowledging the message.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQMessageActions.cs
```diff
@@ -0,0 +1,29 @@
...
+    private readonly IModel channel;
+
+    internal RabbitMQMessageActions(IModel channel)
+    {
+        this.channel = channel;
+    }
+
+    public void BasicReject(ulong deliveryTag)
+    {
+        this.channel.BasicReject(deliveryTag, requeue: false);
```
vivekjilla: just wondering are these BasicReject and BasicAck fire and forget type of calls? Given these are network calls to rabbitmq, do we wait synchronously for the call to complete ( or are there async equivalent methods for these in rabbitmq)

vivekjilla: if we're giving the control to cx to call BasicReject or BasicAck, do we wanna pass the rest of the parameters like requeue, and multiple? Or would they not add value..

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQTriggerBinding.cs
```diff
@@ -31,7 +32,7 @@ internal class RabbitMQTriggerBinding(IRabbitMQService service, string queueName
     public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
     {
         var message = (BasicDeliverEventArgs)value;
-        IReadOnlyDictionary<string, object> bindingData = CreateBindingData(message);
+        IReadOnlyDictionary<string, object> bindingData = CreateBindingData(message, new RabbitMQMessageActions(this.service.Model));
```
vivekjilla: can this break existing apps which don't pass or consume RabbitMQMessageActions?

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
```diff
@@ -54,13 +57,42 @@ public void Verify_BindingDataContract_Values()
...
             Assert.Equal(data[item.Key], item.Value);
         }
     }
+
+    [Fact]
+    public void RabbitMQTriggerAttribute_ManualAck_DefaultsToFalse()
+    {
+        var attribute = new RabbitMQTriggerAttribute("test-queue");
+        bool manualAck = attribute.ManualAck;
+        Assert.False(manualAck, "ManualAck should default to false.");
```
vivekjilla: Can we add more tests for `true` scenario?

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQMessageActions.cs
ishank12: No we dont wait for these calls to complete.  these are fire and forget calls.  There are no async equivalent for this.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQTriggerBinding.cs
ishank12: No is not a breaking change, existing customer code would work as it is.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
ishank12: Log added

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
ishank12: Test `RabbitMQTrigger_ManualAck_BasicAckBehavior` added for behavior of calling basicAck based on manualAck parameter.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQMessageActions.cs
ishank12: done

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -133,10 +142,21 @@ async Task ReceivedHandler(object model, BasicDeliverEventArgs args)
...
+                this.logger.LogDebug($"Acknowledging message for {this.logDetails} since manualAck is not set.");
```
vivekjilla: maybe we can skip the log for default case where manualAck is false?

ishank12: removed..

aloiva:
Wondering if we can reduce the function app code by adding deliveryTag in the messageActions itself as a private member since each execution has a new instance of RabbitMQMessageActions that pertains to the current message. Would make it much easier this way instead of mandating two parameters to make use of this feature:
```
[FunctionName("Function1")]
public void Run([RabbitMQTrigger("test11", ConnectionStringSetting = "rabbitmqconnsetting", ManualAck = true)]string myQueueItem,
    RabbitMQMessageActions messageActions,
    ILogger log)
{
    log.LogInformation($"C# Queue trigger function processed: {myQueueItem}, deliveryTag: {deliveryTag}, messageActions: ");
    
    Console.WriteLine("Acknowleding...");
    messageActions.BasicAck();
    // messageActions.BasicReject(requeue=true);
}```

**Comment by aloiva** [comment]
I see a complication in this implementation. If the customer code acks the message and fails at a later point, the extension tries to reject the already acknowledged message and requeue it as a retry mechanism. Would this throw an error since by definition the message would already be deleted after ack? It would be better to entirely avoid such complicated scenarios by maybe adding a mechanism that tracks if a message has been acked or not.

**Comment by ishank12** [comment]
> Wondering if we can reduce the function app code by adding deliveryTag in the messageActions itself as a private member since each execution has a new instance of RabbitMQMessageActions that pertains to the current message. Would make it much easier this way instead of mandating two parameters to make use of this feature:
> 
> ```
> [FunctionName("Function1")]
> public void Run([RabbitMQTrigger("test11", ConnectionStringSetting = "rabbitmqconnsetting", ManualAck = true)]string myQueueItem,
>     RabbitMQMessageActions messageActions,
>     ILogger log)
> {
>     log.LogInformation($"C# Queue trigger function processed: {myQueueItem}, deliveryTag: {deliveryTag}, messageActions: ");
>     
>     Console.WriteLine("Acknowleding...");
>     messageActions.BasicAck();
>     // messageActions.BasicReject(requeue=true);
> }
> ```

@olalavi , each execution will not have new instance of RabbitMQMessageActions. The channel instance is only one throughout the lifecycle of extension and so is the RabbitMQMessageActions created only once. 
it would be inefficient to create separate instance for each new message.

But it can be done

**Comment by ishank12** [comment]
> I see a complication in this implementation. If the customer code acks the message and fails at a later point, the extension tries to reject the already acknowledged message and requeue it as a retry mechanism. Would this throw an error since by definition the message would already be deleted after ack? It would be better to entirely avoid such complicated scenarios by maybe adding a mechanism that tracks if a message has been tracked or not.

Hey @aloiva ,
Yes that is a good point, 
if BasicAck or Reject happens on the same delivery tag twice in any sequence, it will throw exception on channel and channel will be closed.
I'll try to address this.

File: extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs [hunk 6]
```diff
@@ -62,6 +65,12 @@ public RabbitMQListener(
         // Do not convert the scale-monitor ID to lower-case string since RabbitMQ queue names are case-sensitive.
         this.Descriptor = new ScaleMonitorDescriptor($"{functionId}-RabbitMQTrigger-{queueName}", functionId);
         this.logDetails = $"function: '{functionId}', queue: '{queueName}'";
+
+        // Add a handler to log any errors that occur on the channel.
+        channel.ModelShutdown += (sender, args) =>
+        {
+            logger.LogError($"[!] Channel closed due to error: {args.Exception?.Message}");
```
JatinSanghvi: Are we sure, we are adding this event handler only once to the channel? Want to make sure that we don't log the same error multiple times in case of an error.

JatinSanghvi: Can you end the logging with ` for {this.logDetails}.`?

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQMessageActions.cs
```diff
@@ -0,0 +1,29 @@
...
+    private readonly IModel channel;
+
+    internal RabbitMQMessageActions(IModel channel)
+    {
+        this.channel = channel;
+    }
+
+    public void BasicReject(ulong deliveryTag, bool requeue = false)
+    {
+        this.channel.BasicReject(deliveryTag, requeue: requeue);
```
JatinSanghvi: Just `this.channel.BasicReject(deliveryTag, requeue);` and `this.channel.BasicAck(deliveryTag, multiple);`?

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQMessageActions.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using RabbitMQ.Client;
```
JatinSanghvi, line 6: Remove redundant imports.

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
```diff
@@ -54,13 +63,109 @@ public void Verify_BindingDataContract_Values()
...
```
JatinSanghvi: We should not have tests for trivial checks.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
```diff
@@ -33,6 +33,7 @@ internal sealed class RabbitMQListener : IListener, IScaleMonitor<RabbitMQTrigge
     private readonly ILogger logger;
     private readonly string queueName;
     private readonly ushort prefetchCount;
+    private readonly bool manualAck;
```
JatinSanghvi: We should rename this to `disableAck`. It's entirely on the function whether or not to ack at a later time.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
ishank12: Thanks, this was great catch, I realised that the RabbitMQService is getting cached based on connection string.  If we have multiple connections with same sonnection string, it would cause handler registration multiple times.

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
```diff
@@ -54,13 +63,109 @@ public void Verify_BindingDataContract_Values()
...
+    }
+
+    [Fact]
+    public void RabbitMQTriggerBinding_CreateBindingData_IncludesRabbitMQMessageActions()
+    {
+        ulong deliveryTag = 1;
```
JatinSanghvi: Isn't `Verify_BindingDataContract_Values` already covering this case after your modifications?

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
```diff
@@ -54,13 +63,109 @@ public void Verify_BindingDataContract_Values()
...
+
+        // Find the IBasicConsumer passed to BasicConsume
+        IInvocation basicConsumeInvocation = mockChannel.Invocations
+            .FirstOrDefault(invocation => invocation.Method.Name == "BasicConsume");
+
+        Assert.NotNull(basicConsumeInvocation);
+
+        var consumer = basicConsumeInvocation.Arguments[6] as AsyncEventingBasicConsumer;
+        Assert.NotNull(consumer);
+
```
JatinSanghvi, line 125: Can you explain what `BasicConsume` and `basicConsumeInvocation.Arguments[6]` are and why they are important to test?

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
```diff
@@ -54,13 +63,109 @@ public void Verify_BindingDataContract_Values()
...
+            body: eventArgs.Body.ToArray());
+
+        // Assert
+        if (manualAck)
+        {
+            mockChannel.Verify(channel => channel.BasicAck(It.IsAny<ulong>(), It.IsAny<bool>()), Times.Never, "BasicAck should not be called when ManualAck is true.");
+        }
+        else
+        {
+            mockChannel.Verify(channel => channel.BasicAck(It.IsAny<ulong>(), It.IsAny<bool>()), Times.AtLeastOnce, "BasicAck should be called when ManualAck is false.");
```
JatinSanghvi: Change `AtLeastOnce` to `Once` since the mock is returning success unconditionally and we only want to ack once in this case.

JatinSanghvi (changes requested):

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
JatinSanghvi: Open to discussion over this. Other candidates can be `disableAckOnSuccess` and `disableAutoAck`.

aloiva:
> Yes that is a good point,
if BasicAck or Reject happens on the same delivery tag twice in any sequence, it will throw exception on channel and channel will be closed.
I'll try to address this.

@ishank12 , another scenario to address would be where manualAck is set to false (or disableAck is set to true) and yet the customer makes use of MessageActions to ack or reject in their function code. Extension will also try to do the same after function execution and will throw an error. Instead of the methods in MessageActions being independent, making them dependent on manualAck/disableAck bool might fix this. Or making how MessageActions is received by the function code based on manualAck/disableAck (maybe return null in such cases instead of always returning the initialised object?) Just throwing out ideas here..

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
ishank12: > Can you end the logging with ` for {this.logDetails}.`?

This channel closure is not consumer specific. If this the channel itself would be closed and no trigger would work.  This is error condition and should not happen anytime.

extension/WebJobs.Extensions.RabbitMQ/Trigger/RabbitMQListener.cs
ishank12: disableAck seems good.

extension/WebJobs.Extensions.RabbitMQ.Tests/RabbitMQTriggerBindingTests.cs
ishank12: changed to this one

```
// Find the Consumer instance passed to RabbitMQService.Consume method
IInvocation consumeInvocation = mockservice.Invocations
    .FirstOrDefault(invocation => invocation.Method.Name == "Consume");

Assert.NotNull(consumeInvocation);

var consumer = consumeInvocation.Arguments[2] as AsyncEventingBasicConsumer;
```

_AsyncEventingBasicConsumer is the third parameter of Consume method_

ishank12: Basically getting the exact consumer instance created inside startAsync method and explicitly delivering a message to check if BasicAck method is called or not.

JatinSanghvi (approved):

aloiva (approved):

